PR #56, opened 2024-03-20
Test: oracle daemon config

TheDZhon:
The PR contains a unit test suite shielding the `OracleDaemonConfig` contract.

Find the attached coverage report below :star_struck: 

---

image


test/0.8.9/oracleDaemonConfig.test.ts
```diff
@@ -0,0 +1,279 @@
...
+          .to.emit(oracleDaemonConfig, "ConfigValueSet")
+          .withArgs("key2", "0xdead");
+
+        await expect(oracleDaemonConfig.getList(["unknownKey"]))
+          .to.be.revertedWithCustomError(oracleDaemonConfig, "ValueDoesntExist")
+          .withArgs("unknownKey");
+      });
+    });
+
+    it("getList works for empty keys list", async () => {
```
tamtamchik, line 252: Breaks test naming, may be just "Works..." (same in test below)

test/0.8.9/oracleDaemonConfig.test.ts
```diff
@@ -0,0 +1,279 @@
...
+        .withArgs("", defaultValue);
+
+      await expect(oracleDaemonConfig.connect(configPrimaryManager).unset(""))
+        .to.emit(oracleDaemonConfig, "ConfigValueUnset")
+        .withArgs("");
+    });
+  });
+
+  context("Function `getList`", () => {
+    context("Reverts", () => {
```
tamtamchik, line 236: Do we need this duplication? I don't see the problem this isolation solves. The "Reverts" part is included in `it(...)` block, anyway.

test/0.8.9/oracleDaemonConfig.test.ts
```diff
@@ -0,0 +1,279 @@
...
+      expect(await newODCInstance.hasRole(adminRole, admin)).to.equal(true);
+
+      const configManagerRole = await oracleDaemonConfig.CONFIG_MANAGER_ROLE();
+      expect(await newODCInstance.getRoleMemberCount(configManagerRole)).to.equal(0);
+      expect(await newODCInstance.hasRole(configManagerRole, configPrimaryManager)).to.equal(false);
+      expect(await newODCInstance.hasRole(configManagerRole, configSecondaryManager)).to.equal(false);
+    });
+  });
+
+  context("Set and get values", () => {
```
tamtamchik, line 78: Split to set and get contexts? Then, there is no need to mention the function again inside the `it(...)` block.

test/0.8.9/oracleDaemonConfig.test.ts
```diff
@@ -0,0 +1,279 @@
...
+    it("Get works after set with empty key", async () => {
+      await expect(oracleDaemonConfig.connect(configPrimaryManager).set("", defaultValue))
+        .to.emit(oracleDaemonConfig, "ConfigValueSet")
+        .withArgs("", defaultValue);
+
+      expect(await oracleDaemonConfig.get("")).to.equal(defaultValue);
+    });
+  });
+
+  context("Function `update`", () => {
```
tamtamchik, line 129: By default, we test functions, so do we need to highlight that it is a function? In other words, `update` here will work just fine (combined with breaking context in above)

tamtamchik (approved):
I have some concerns regarding naming test contexts and test cases.

Otherwise, it looks great! 🎉 ✅

test/0.8.9/oracleDaemonConfig.test.ts
TheDZhon: Fixed in 656b32d83a9d8b53e6eca5a10218a706d9ab5c76, thanks!

test/0.8.9/oracleDaemonConfig.test.ts
TheDZhon: It looks like a way to outline the negative tests surface more clearly to me :thinking:

test/0.8.9/oracleDaemonConfig.test.ts
TheDZhon: Yep, fixed in 656b32d83a9d8b53e6eca5a10218a706d9ab5c76, thanks!

test/0.8.9/oracleDaemonConfig.test.ts
TheDZhon: Agree, changed this in 656b32d83a9d8b53e6eca5a10218a706d9ab5c76 :heart:

mymphe:
Thank you! Approved!

Nitpick. I'd avoid using `Works` in the test description and write out what the function actually does.

mymphe (approved):

PR #59, opened 2024-03-21
Feature: Multiprover contract

vp42ldo:
Feature: Multiprover contract according to https://hackmd.io/cGl49k1wTVelrKUW0DhAIw?view#

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,3 +16,4 @@ coverage.json
 /foundry/out
 
 .env
+.DS_Store
--- a/contracts/0.8.9/oracle/Multiprover.sol
+++ b/contracts/0.8.9/oracle/Multiprover.sol
@@ -1,0 +1,212 @@
+// SPDX-FileCopyrightText: 2024 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
+pragma solidity 0.8.9;
+
+import { AccessControlEnumerable } from "../utils/access/AccessControlEnumerable.sol";
+import { SafeCast } from "@openzeppelin/contracts-v4.4/utils/math/SafeCast.sol";
+
+interface ILidoZKOracle {
+    function getReport(uint256 refSlot) external view returns  (
+        bool success,
+        uint256 clBalanceGwei,
+        uint256 numValidators,
+        uint256 exitedValidators
+	);
+}
+
+contract Multiprover is ILidoZKOracle, AccessControlEnumerable {
+    using SafeCast for uint256;
+
+    error AdminCannotBeZero();
+
+    // zk Oracles commetee
+    error DuplicateMember();
+    error NonMember();
+    error QuorumTooSmall(uint256 minQuorum, uint256 receivedQuorum);
+    error AddressCannotBeZero();
+
+    error NoConsensus();
+
+    event MemberAdded(address indexed addr, uint256 newTotalMembers, uint256 newQuorum);
+    event MemberRemoved(address indexed addr, uint256 newTotalMembers, uint256 newQuorum);
+    event QuorumSet(uint256 newQuorum, uint256 totalMembers, uint256 prevQuorum);
+
+    /// @notice An ACL role granting the permission to modify members list members and
+    /// change the quorum by calling addMember, removeMember, and setQuorum functions.
+    bytes32 public constant MANAGE_MEMBERS_AND_QUORUM_ROLE =
+        keccak256("MANAGE_MEMBERS_AND_QUORUM_ROLE");
+
+    /// @dev A quorum value that effectively disables the oracle.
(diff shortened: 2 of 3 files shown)


contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -562,14 +619,85 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        uint256 _postCLValidators,
+        uint256 _reportTimestamp
+    ) internal {
         if (_preCLBalance <= _unifiedPostCLBalance) return;
         uint256 oneOffCLBalanceDecreaseBP = (MAX_BASIS_POINTS * (_preCLBalance - _unifiedPostCLBalance)) /
             _preCLBalance;
         if (oneOffCLBalanceDecreaseBP > _limitsList.oneOffCLBalanceDecreaseBPLimit) {
             revert IncorrectCLBalanceDecrease(oneOffCLBalanceDecreaseBP);
         }
+        _checkAccountingReportZKP(_unifiedPostCLBalance, _postCLValidators, _reportTimestamp);
```
TheDZhon, line 632: should it be inside the if above?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -415,7 +471,7 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         uint256 _sharesRequestedToBurn,
         uint256 _preCLValidators,
         uint256 _postCLValidators
-    ) external view {
+    ) external {
```
TheDZhon, line 474: why do we lift mutability?

contracts/0.8.9/sanity_checks/SanityFuse.sol
TheDZhon, line 1: not sure if we need this

TheDZhon (changes requested):
:+1: Had a first glance, looks pretty promising

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp42ldo: Because of the sanity fuse. It needs to know was the report successful or not for it's logic

contracts/0.8.9/sanity_checks/SanityFuse.sol
vp42ldo: that would simplify things a lot

contracts/0.8.9/oracle/Tiebreaker.sol
```diff
@@ -0,0 +1,45 @@
...
+pragma solidity 0.8.9;
+
+import { AccessControlEnumerable } from "../utils/access/AccessControlEnumerable.sol";
+import { SafeCast } from "@openzeppelin/contracts-v4.4/utils/math/SafeCast.sol";
+import { ILidoZKOracle } from "./ILidoZKOracle.sol";
+
+contract Tiebreaker is ILidoZKOracle, AccessControlEnumerable {
+    using SafeCast for uint256;
+
+    // TODO: separate role to submit reports
```
TheDZhon, line 12:
```suggestion
    /// @notice An ACL role granting the permission to submit the data for a tiebraker report.
    bytes32 public constant TIE_BRAKER_SUBMIT_ROLE = keccak256("TIE_BRAKER_SUBMIT_ROLE");
```

contracts/0.8.9/oracle/Tiebreaker.sol
```diff
@@ -0,0 +1,45 @@
...
+        onlyRole(DEFAULT_ADMIN_ROLE) {
+        _reports[refSlot] = Report(success, clBalanceGwei, numValidators, exitedValidators);
+    }
+
+    function removeReport(uint256 refSlot) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        delete _reports[refSlot];
+    }
+
+    function getReport(uint256 refSlot) external view override returns (bool, uint256, uint256, uint256) {
+        Report memory report = _reports[refSlot];
```
TheDZhon, line 41: It will revert if no report is present, while it should return `success == false` instead

TheDZhon (commented):
A few more things

TheDZhon:
Mind the gap: decided to not review the tests for now :warning:

contracts/0.8.9/oracle/Tiebreaker.sol
vp42ldo: it's mapping, not an array. So I suppose that should return a default value for the value type. In the case of a struct like Report, each field in the struct will be initialized to its default value. So, the success should be initialized as false.

PR #67, opened 2024-04-05
Negative rebase limit

vp4242:
Negative rebase limit


contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -34,6 +35,20 @@ interface IWithdrawalQueue {
         returns (WithdrawalRequestStatus[] memory statuses);
 }
 
+interface ILidoBaseOracle {
+    function SECONDS_PER_SLOT() external view returns (uint256);
+    function GENESIS_TIME() external view returns (uint256);
+}
+
+interface ILidoZKOracle {
```
TheDZhon, line 43:
```suggestion
interface ILidoCLStateOracle {
```

test/0.8.9/contracts/OracleReportSanityCheckerWrapper.sol
```diff
@@ -0,0 +1,25 @@
+// SPDX-FileCopyrightText: 2024 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
```
TheDZhon, line 2:
```suggestion
// SPDX-FileCopyrightText: 2024 Lido <info@lido.fi>
// SPDX-License-Identifier: GPL-3.0
// NB: for testing purposes only
```

test/0.8.9/contracts/ZkOracleMock.sol
```diff
@@ -0,0 +1,39 @@
+// SPDX-FileCopyrightText: 2024 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
```
TheDZhon, line 2:
```suggestion
// SPDX-FileCopyrightText: 2024 Lido <info@lido.fi>
// SPDX-License-Identifier: GPL-3.0
// NB: for testing purposes only
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -430,7 +469,8 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         _checkSharesRequestedToBurn(_sharesRequestedToBurn);
 
         // 4. Consensus Layer one-off balance decrease
-        _checkOneOffCLBalanceDecrease(limitsList, _preCLBalance, _postCLBalance + _withdrawalVaultBalance);
+        _checkOneOffCLBalanceDecrease(limitsList, _preCLBalance,
```
TheDZhon, line 472: I would rather rename this method tbh

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -331,6 +355,17 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         _updateLimits(limitsList);
     }
 
+    /// @notice Returns the address of the negative rebase oracle
+    function getNegativeRebaseOracle() public view returns (address) {
+        return _negativeRebaseOracle;
+    }
+
+    /// @notice Sets the address of the negative rebase oracle
+    /// @param negativeRebaseOracle address of the negative rebase oracle
```
TheDZhon, line 364: may add a remark that the address might be zero

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -331,6 +355,17 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         _updateLimits(limitsList);
     }
 
+    /// @notice Returns the address of the negative rebase oracle
+    function getNegativeRebaseOracle() public view returns (address) {
```
TheDZhon, line 359:
```suggestion
    function getCLStateOracle() public view returns (address) {
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -331,6 +355,17 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        return _negativeRebaseOracle;
+    }
+
+    /// @notice Sets the address of the negative rebase oracle
+    /// @param negativeRebaseOracle address of the negative rebase oracle
+    function setNegativeRebaseOracle(address negativeRebaseOracle) external onlyRole(DEFAULT_ADMIN_ROLE) {
```
TheDZhon, line 365:
```suggestion
    function setCLStateOracle(address clStateOracle) external onlyRole(DEFAULT_ADMIN_ROLE) {
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -92,6 +107,11 @@ struct LimitsListPacked {
     uint64 maxPositiveTokenRebase;
 }
 
+struct RebaseData {
```
TheDZhon, line 110:
```suggestion
struct CLRebaseData {
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -92,6 +107,11 @@ struct LimitsListPacked {
...
+    uint64 rebaseValue;
```
TheDZhon, line 111:
```suggestion
    uint192 clRebaseValue;
```

as Solidity/EVM use 32-bytes long machine words, it doesn't make sense to make struct fields narrower than that

TheDZhon (changes requested):
:+1: It's an excellent way to push things closer to Mainnet.

I suggest considering my comments and findings to improve the code and figure out the plan forward with audits :goggles:

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -562,14 +602,82 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
         uint256 _preCLBalance,
-        uint256 _unifiedPostCLBalance
-    ) internal pure {
+        uint256 _unifiedPostCLBalance,
+        uint256 _reportTimestamp
+    ) internal {
         if (_preCLBalance <= _unifiedPostCLBalance) return;
         uint256 oneOffCLBalanceDecreaseBP = (MAX_BASIS_POINTS * (_preCLBalance - _unifiedPostCLBalance)) /
             _preCLBalance;
         if (oneOffCLBalanceDecreaseBP > _limitsList.oneOffCLBalanceDecreaseBPLimit) {
```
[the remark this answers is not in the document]
vp42ldo: that's a good question. Probably not as the new one will prevail. I will remove it then

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp42ldo: unless we would like to pack it. I guess we won't in this case

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -238,14 +275,36 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
     {
         LimitsList memory limitsList = _limits.unpack();
-        limitsList.oneOffCLBalanceDecreaseBPLimit = _oneOffCLBalanceDecreaseBPLimit;
+        limitsList.cLBalanceDecreaseHoursSpan = _cLBalanceDecreaseHoursSpan;
+        _updateLimits(limitsList);
+    }
+
+    /// @notice Sets the new value for the cLBalanceOraclesDiffBPLimit
+    /// @param _cLBalanceOraclesDiffBPLimit new cLBalanceOraclesDiffBPLimit value
+    function setCLBalanceOraclesDiffBPLimit(uint256 _cLBalanceOraclesDiffBPLimit)
```
folkyatina, line 302: Error Margin?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +636,71 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         }
     }
 
-    function _checkOneOffCLBalanceDecrease(
+    function _addRebaseValue(uint256 _value, uint256 _timestamp) internal {
+        _rebaseData.push(CLRebaseData(SafeCast192.toUint192(_value), SafeCast.toUint64(_timestamp)));
+    }
+
+    function sumRebaseValuesNotOlderThan(uint256 _timestamp) public view returns (uint256) {
```
folkyatina, line 643: Is this intentionally left public? If so, it requires to be placed upper in the file and some NatSpec should be added

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +636,71 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
-        if (oneOffCLBalanceDecreaseBP > _limitsList.oneOffCLBalanceDecreaseBPLimit) {
-            revert IncorrectCLBalanceDecrease(oneOffCLBalanceDecreaseBP);
+
+        _addRebaseValue(_preCLBalance - _unifiedPostCLBalance, _reportTimestamp);
+
+        uint256 pastTimestamp = _reportTimestamp - _limitsList.cLBalanceDecreaseHoursSpan * 1 hours;
+        uint256 rebaseSum = sumRebaseValuesNotOlderThan(pastTimestamp);
+
+        uint256 balanceDiffBP = MAX_BASIS_POINTS * rebaseSum / (_unifiedPostCLBalance + rebaseSum);
+        // NOTE: Base points is 10_000, so 320 BP is 3.20%
```
folkyatina, line 671: Comment is irrelevant

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +636,71 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
-            _preCLBalance;
-        if (oneOffCLBalanceDecreaseBP > _limitsList.oneOffCLBalanceDecreaseBPLimit) {
-            revert IncorrectCLBalanceDecrease(oneOffCLBalanceDecreaseBP);
+
+        _addRebaseValue(_preCLBalance - _unifiedPostCLBalance, _reportTimestamp);
+
+        uint256 pastTimestamp = _reportTimestamp - _limitsList.cLBalanceDecreaseHoursSpan * 1 hours;
+        uint256 rebaseSum = sumRebaseValuesNotOlderThan(pastTimestamp);
+
+        uint256 balanceDiffBP = MAX_BASIS_POINTS * rebaseSum / (_unifiedPostCLBalance + rebaseSum);
```
folkyatina, line 670: Loss of precision because of int division can influence the comparison later. Maybe redo the calculations to get rid of the division?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +636,71 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        if (balanceDiffBP <= _limitsList.cLBalanceDecreaseBPLimit) {
+            // If the diff is less than limit we are finishing check
+            return;
+        }
+
+        address clStateOracle = getCLStateOracle();
+        // If there is no negative rebase oracle, then we don't need to check the zk report
+        if (clStateOracle == address(0)) {
+            // If there is no oracle and the diff is more than limit, we revert
+            revert IncorrectCLBalanceDecreaseForSpan(balanceDiffBP);
```
folkyatina, line 681: The error name looks like it should be a span parameter also.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +636,71 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+        (bool success, uint256 clOracleBalanceGwei,,)
+            = ILidoCLStateOracle(clStateOracle).getReport(refSlot);
+
+        if (success) {
+            uint256 clBalanceWei = clOracleBalanceGwei * 1 gwei;
+            uint256 balanceDiff = (clBalanceWei > _unifiedPostCLBalance) ?
+                clBalanceWei - _unifiedPostCLBalance : _unifiedPostCLBalance - clBalanceWei;
+            uint256 balanceDifferenceBP = MAX_BASIS_POINTS * balanceDiff / clBalanceWei;
+            // NOTE: Base points is 10_000, so 74 BP is 0.74%
```
folkyatina, line 697: irrelevant comment

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +636,71 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+            if (balanceDifferenceBP >= _limitsList.cLBalanceOraclesDiffBPLimit) {
```
folkyatina, line 698: This comparison does not suffer from precision loss, but maybe it's worth making it uniformly.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +636,71 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        address clStateOracle = getCLStateOracle();
+        // If there is no negative rebase oracle, then we don't need to check the zk report
+        if (clStateOracle == address(0)) {
+            // If there is no oracle and the diff is more than limit, we revert
+            revert IncorrectCLBalanceDecreaseForSpan(balanceDiffBP);
+        }
+
+        address accountingOracle = ILidoLocator(getLidoLocator()).accountingOracle();
+        uint256 refSlot = (_reportTimestamp -
+            ILidoBaseOracle(accountingOracle).GENESIS_TIME()) /
```
folkyatina, line 686: It looks like GENESIS_TIME and SECONDS_PER_SLOT can be immutable contract members initialized in the constructor because they do not change after deployment, only from chain to chain.

folkyatina (approved):
Thanks for your work. Minor suggestions for the contract.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: ?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: thanks. I don't mind making it internal but it seems that public is more transparent

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
folkyatina: Just a suggestion for better naming

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -34,6 +35,20 @@ interface IWithdrawalQueue {
         returns (WithdrawalRequestStatus[] memory statuses);
 }
 
+interface ILidoBaseOracle {
```
TheDZhon, line 38:
```suggestion
interface IBaseOracle {
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -34,6 +35,20 @@ interface IWithdrawalQueue {
...
+    function GENESIS_TIME() external view returns (uint256);
+}
+
+interface ILidoCLStateOracle {
+    function getReport(uint256 refSlot) external view returns  (
+        bool success,
+        uint256 clBalanceGwei,
+        uint256 numValidators,
+        uint256 exitedValidators
+	);
```
TheDZhon, line 49: please check the formatting

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -44,10 +59,19 @@ struct LimitsList {
...
+    ///     a timespan of N hours
+    /// @dev Represented in the Basis Points (100% == 10_000)
+    uint256 cLBalanceDecreaseBPLimit;
+
+    /// @notice The timespan of the max decrease of the total validators' balances on the Consensus Layer
+    /// @dev Represented in hours
+    uint256 cLBalanceDecreaseHoursSpan;
+
+    /// @notice The maximum divergence between the total balances of validators on the Consensus Layer
+    ///     as reported by the Oracle and the state of the Consensus Layer Oracle.
```
TheDZhon, line 72:
```suggestion
    ///     as reported by the AccountingOracle and the state of the Consensus Layer Oracle.
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -82,7 +106,9 @@ struct LimitsList {
 /// @dev The packed version of the LimitsList struct to be effectively persisted in storage
 struct LimitsListPacked {
```
TheDZhon, line 107: let's try to find some space to make this struct occupying only a single storage slot :warning: 

now it's 272 bits while we can shorten some fields to make it ≤256, e.g. make `requestTimestampMargin` as `uint48`

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -121,19 +152,27 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
 
     ILidoLocator private immutable LIDO_LOCATOR;
+    uint256 private immutable GENESIS_TIME;
+    uint256 private immutable SECONDS_PER_SLOT;
 
+    address private _clStateOracle;
     LimitsListPacked private _limits;
 
+    /// @dev The array of the rebase values and the corresponding timestamps
+    CLRebaseData[] private _rebaseData;
```
TheDZhon, line 170:
```suggestion
    CLRebaseData[] private _clRebases;
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -238,14 +281,36 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
     {
         LimitsList memory limitsList = _limits.unpack();
-        limitsList.oneOffCLBalanceDecreaseBPLimit = _oneOffCLBalanceDecreaseBPLimit;
+        limitsList.cLBalanceDecreaseBPLimit = _cLBalanceDecreaseBPLimit;
+        _updateLimits(limitsList);
+    }
+
+    /// @notice Sets the new value for the cLBalanceDecreaseHoursSpan
+    /// @param _cLBalanceDecreaseHoursSpan new cLBalanceDecreaseHoursSpan value
+    function setCLBalanceDecreaseHoursSpan(uint256 _cLBalanceDecreaseHoursSpan)
```
TheDZhon, line 297: I am not entirely convinced if we need to change both values separately and not via a single method, wdyt?

e.g.:

```solidity
function setCLBalanceDecreaseLimit(uint256 _clBalanceDrecreaseBPLimit, uint256 _clBalanceDecreaseHoursSpan)
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -238,14 +281,36 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        onlyRole(CL_BALANCE_DECREASE_LIMIT_MANAGER_ROLE)
+    {
+        LimitsList memory limitsList = _limits.unpack();
+        limitsList.cLBalanceDecreaseHoursSpan = _cLBalanceDecreaseHoursSpan;
+        _updateLimits(limitsList);
+    }
+
+    /// @notice Sets the new value for the cLBalanceOraclesErrorMarginBPLimit
+    /// @param _cLBalanceOraclesErrorMarginBPLimit new cLBalanceOraclesErrorMarginBPLimit value
+    function setCLBalanceOraclesErrorMarginBPLimit(uint256 _cLBalanceOraclesErrorMarginBPLimit)
```
TheDZhon, line 308: it might be needed to set together with the cl oracle :thinking:

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -533,6 +616,23 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         );
     }
 
+    /// @notice Returns the sum of the rebase values not older than the provided timestamp
+    /// @param _timestamp the timestamp to check the rebase values
+    /// @return rebaseValuesSum the sum of the rebase values not older than the provided timestamp
+    function sumRebaseValuesNotOlderThan(uint256 _timestamp) public view returns (uint256) {
+        uint256 rebaseValuesSum = 0;
+        int256 slot = int256(_rebaseData.length) - 1;
+        while (slot >= 0) {
```
TheDZhon, line 625: Why not old plain ~soviet~ for loop?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -533,6 +616,23 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         );
     }
 
+    /// @notice Returns the sum of the rebase values not older than the provided timestamp
+    /// @param _timestamp the timestamp to check the rebase values
+    /// @return rebaseValuesSum the sum of the rebase values not older than the provided timestamp
+    function sumRebaseValuesNotOlderThan(uint256 _timestamp) public view returns (uint256) {
+        uint256 rebaseValuesSum = 0;
```
TheDZhon, line 623:
```suggestion
        uint256 rebaseValuesSum;
```

gas golfing, haha (disregard if you find yourself uncomfortable with this change)

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +659,54 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         }
     }
 
-    function _checkOneOffCLBalanceDecrease(
+    function _addRebaseValue(uint256 _value, uint256 _timestamp) internal {
+        _rebaseData.push(CLRebaseData(SafeCast192.toUint192(_value), SafeCast.toUint64(_timestamp)));
```
TheDZhon, line 663: shall we enforce invariants for this data?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -791,3 +948,31 @@ library LimitsListUnpacker {
         res.maxNodeOperatorsPerExtraDataItemCount = _limitsList.maxNodeOperatorsPerExtraDataItemCount;
     }
 }
+
+// OpenZeppelin Contracts (last updated v5.0.2) (utils/math/SafeCast.sol)
+// Code extracted from https://github.com/OpenZeppelin/openzeppelin-contracts/releases/tag/v5.0.2
+
+library SafeCast192 {
```
TheDZhon, line 955: Let's extract this to a separate file or not introduce a library at all

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -533,6 +616,23 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         );
     }
 
+    /// @notice Returns the sum of the rebase values not older than the provided timestamp
+    /// @param _timestamp the timestamp to check the rebase values
+    /// @return rebaseValuesSum the sum of the rebase values not older than the provided timestamp
+    function sumRebaseValuesNotOlderThan(uint256 _timestamp) public view returns (uint256) {
+        uint256 rebaseValuesSum = 0;
+        int256 slot = int256(_rebaseData.length) - 1;
```
TheDZhon, line 624:
```suggestion
        int256 clRebaseIndex = int256(_rebaseData.length) - 1;
```

:warning: let's avoid using the `slot` term here because it's a bit ambiguous (it might be confused with the Consensus Layer slots)

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
TheDZhon: e.g.:

```solidity
for(int256 clRebaseIndex = int256(_rebaseData.length) - 1; clRebaseIndex >= 0; --clRebaseIndex)
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +659,54 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+    function _checkCLBalanceDecrease(
         LimitsList memory _limitsList,
         uint256 _preCLBalance,
-        uint256 _unifiedPostCLBalance
-    ) internal pure {
+        uint256 _unifiedPostCLBalance,
+        uint256 _reportTimestamp
+    ) internal {
         if (_preCLBalance <= _unifiedPostCLBalance) return;
```
TheDZhon, line 672: I would document this early return tbh

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +659,54 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
-        uint256 oneOffCLBalanceDecreaseBP = (MAX_BASIS_POINTS * (_preCLBalance - _unifiedPostCLBalance)) /
-            _preCLBalance;
-        if (oneOffCLBalanceDecreaseBP > _limitsList.oneOffCLBalanceDecreaseBPLimit) {
-            revert IncorrectCLBalanceDecrease(oneOffCLBalanceDecreaseBP);
+
+        _addRebaseValue(_preCLBalance - _unifiedPostCLBalance, _reportTimestamp);
```
TheDZhon, line 674:
```suggestion
        _addNegativeCLRebase(_preCLBalance - _unifiedPostCLBalance, _reportTimestamp);
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +659,54 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+        uint256 pastTimestamp = _reportTimestamp - _limitsList.cLBalanceDecreaseHoursSpan * 1 hours;
+        uint256 rebaseSum = sumRebaseValuesNotOlderThan(pastTimestamp);
```
TheDZhon, line 677: It reads like we sum up every rebase, while looking only for negative ones :warning:

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +659,54 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+        uint256 rebaseSumBP = MAX_BASIS_POINTS * rebaseSum;
+        uint256 limitMulByStartBalance = _limitsList.cLBalanceDecreaseBPLimit * (_unifiedPostCLBalance + rebaseSum);
+        if (rebaseSumBP < limitMulByStartBalance) {
+            // If the diff is less than limit we are finishing check
+            return;
+        }
+
+        address clStateOracle = getCLStateOracle();
+        // If there is no negative rebase oracle, then we don't need to check the zk report
```
TheDZhon, line 687: Why 'zk' report? usage of zk technology for this oracle shouldn't be taken for granted

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +659,54 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+            uint256 clBalanceWei = clOracleBalanceGwei * 1 gwei;
+            uint256 balanceDiff = (clBalanceWei > _unifiedPostCLBalance) ?
+                clBalanceWei - _unifiedPostCLBalance : _unifiedPostCLBalance - clBalanceWei;
+            uint256 balanceDifferenceBP = MAX_BASIS_POINTS * balanceDiff / clBalanceWei;
+            if (balanceDifferenceBP >= _limitsList.cLBalanceOraclesErrorMarginBPLimit) {
+                revert ClBalanceMismatch(_unifiedPostCLBalance, clBalanceWei);
+            }
+            emit ConfirmNegativeRebase(refSlot, clBalanceWei);
+        } else {
+            revert CLStateReportIsNotReady();
```
TheDZhon, line 709: Should this error have a bit different name to say something like '**can't verify negative rebase** because of the missing cl oracle report'?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +659,54 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+        if (success) {
+            uint256 clBalanceWei = clOracleBalanceGwei * 1 gwei;
+            uint256 balanceDiff = (clBalanceWei > _unifiedPostCLBalance) ?
+                clBalanceWei - _unifiedPostCLBalance : _unifiedPostCLBalance - clBalanceWei;
+            uint256 balanceDifferenceBP = MAX_BASIS_POINTS * balanceDiff / clBalanceWei;
+            if (balanceDifferenceBP >= _limitsList.cLBalanceOraclesErrorMarginBPLimit) {
+                revert ClBalanceMismatch(_unifiedPostCLBalance, clBalanceWei);
+            }
+            emit ConfirmNegativeRebase(refSlot, clBalanceWei);
```
TheDZhon, line 707:
```suggestion
            emit NegativeRebaseConfirmed(refSlot, clBalanceWei);
```

events usually use past tense

TheDZhon (changes requested):
Thank you for polishing up this project further :star_struck: 

I'd recommend to revise some of the approaches used, see the review comments :pray:

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: Pros: less functions
Cons: may accidentally change other value

🤷‍♂️

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: that probably make sense since there should be direct deps on each other

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: those values are used together within the check, so it's probably worth checking that they aligned

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: In that case it's probably worth changing other error to confirm this naming logic

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
TheDZhon: yep

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
TheDZhon: > Cons: may accidentally change other value
For me t's vise versa: way more easier to change one value and accidentally forget about their coupling

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +643,55 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        _clRebases.push(CLRebaseData(SafeCastExt.toUint192(_value), SafeCast.toUint64(_timestamp)));
+    }
+
+    function _checkCLBalanceDecrease(
         LimitsList memory _limitsList,
         uint256 _preCLBalance,
-        uint256 _unifiedPostCLBalance
-    ) internal pure {
+        uint256 _unifiedPostCLBalance,
+        uint256 _reportTimestamp
```
avsetsin, line 654: ⚠️ This method is expected `_reportTimestamp` as an argument, but it's called from `checkAccountingOracleReport` with `_timeElapsed`: https://github.com/lidofinance/core/blob/e2c8ba8c785d552751b47d5240c7cbe8ea4802bd/contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol#L500-L502

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +646,55 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        uint256 rebaseSum = sumNegativeRebasesNotOlderThan(pastTimestamp);
+
+        uint256 rebaseSumBP = MAX_BASIS_POINTS * rebaseSum;
+        uint256 limitMulByStartBalance = _limitsList.cLBalanceDecreaseBPLimit * (_unifiedPostCLBalance + rebaseSum);
+        if (rebaseSumBP < limitMulByStartBalance) {
+            // If the diff is less than limit we are finishing check
+            return;
+        }
+
+        address clStateOracle = getCLStateOracle();
```
avsetsin, line 674: I would suggest to consider a more flexible implementation. Right now it seems that the interface of the second opinion is mainly dictated by the implementation of the ZKP oracle.

A more flexible option might be a push model rather than a pool model. SanityChecker, in case some event occurs (in our case rebase below treshold), could ask for some second opinion contract by sending the values of the current report. Smth like:

```solidity
interface SecondOpinion {
    wdyt(
        uint256 _timeElapsed,
        uint256 _preCLBalance,
        uint256 _postCLBalance,
        uint256 _withdrawalVaultBalance,
        uint256 _preCLValidators,
        uint256 _postCLValidators
    ) external;
}
```

The second opinion contract in this case could be quite variable:
- check against the ZKP oracle report
- check against beacon root proof of state.slashings or inactivity leak
- check against approved values by a multisig

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +646,55 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+        (bool success, uint256 clOracleBalanceGwei,,)
+            = ILidoCLStateOracle(clStateOracle).getReport(refSlot);
+
+        if (success) {
+            uint256 clBalanceWei = clOracleBalanceGwei * 1 gwei;
+            uint256 balanceDiff = (clBalanceWei > _unifiedPostCLBalance) ?
+                clBalanceWei - _unifiedPostCLBalance : _unifiedPostCLBalance - clBalanceWei;
+            uint256 balanceDifferenceBP = MAX_BASIS_POINTS * balanceDiff / clBalanceWei;
+            if (balanceDifferenceBP >= _limitsList.cLBalanceOraclesErrorMarginBPLimit) {
```
avsetsin, line 692: Abstraction leaked. Sanity check contract knows that ZKP oracle has some peculiarities. I guess it's better to handle this on ZKP oracle contract side

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +646,55 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        uint256 refSlot = (_reportTimestamp - GENESIS_TIME) / SECONDS_PER_SLOT;
+
+        (bool success, uint256 clOracleBalanceGwei,,)
+            = ILidoCLStateOracle(clStateOracle).getReport(refSlot);
+
+        if (success) {
+            uint256 clBalanceWei = clOracleBalanceGwei * 1 gwei;
+            uint256 balanceDiff = (clBalanceWei > _unifiedPostCLBalance) ?
+                clBalanceWei - _unifiedPostCLBalance : _unifiedPostCLBalance - clBalanceWei;
+            uint256 balanceDifferenceBP = MAX_BASIS_POINTS * balanceDiff / clBalanceWei;
```
avsetsin, line 691: ⚠️ Division by 0 can happen here when there are no validators left on CL

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: in the current design it's not possible but I see your point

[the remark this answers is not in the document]
vp4242: in other case it was critical about precision. here it'is not but considering div by 0 case — worth changing as well

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -533,6 +608,21 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
     }
 
+    /// @notice Returns the sum of the negative rebase values not older than the provided timestamp
+    /// @param _timestamp the timestamp to check the rebase values
+    /// @return rebaseValuesSum the sum of the negative rebase values not older than the provided timestamp
+    function sumNegativeRebasesNotOlderThan(uint256 _timestamp) public view returns (uint256) {
+        uint256 sum;
+        for (int256 index = int256(_clNegativeRebases.length) - 1; index >= 0; index--) {
+            if (_clNegativeRebases[uint256(index)].timestamp >= SafeCast.toUint64(_timestamp)) {
+                sum += _clNegativeRebases[uint256(index)].value;
```
avsetsin, line 618: I think there is an issue if you operate with absolute values: https://hackmd.io/@lido/HkRweE9Z0

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -618,14 +774,14 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         LimitsList memory _limitsList,
         address _withdrawalQueue,
         uint256 _lastFinalizableId,
-        uint256 _reportTimestamp
+        uint256 reportTimestamp
```
TheDZhon, line 777: should it be rolled back?

TheDZhon (changes requested):
LGTM except only for a few improvements and edits

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -77,23 +107,46 @@ struct LimitsList {
...
+    uint48 requestTimestampMargin;
     uint64 maxPositiveTokenRebase;
+    uint16 initialSlashingAmountPWei;
+    uint16 inactivityPenaltiesAmountPWei;
+    uint16 clBalanceOraclesErrorUpperBPLimit;
+}
+
+struct ReportData {
+    uint64 timestamp;
+    uint64 exitedValidatorsCount;
```
TheDZhon, line 143: nitpick: just to be compat with the `StakingModuleSummary` struct naming.

```suggestion
    uint64 totalExitedValidators;
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -77,23 +107,46 @@ struct LimitsList {
...
+    uint128 negativeCLRebase;
```
TheDZhon, line 144: nitpick: just to avoid future confusion

```suggestion
    uint128 negativeCLRebaseWei;
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +633,96 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+            stakingRouterExitedValidators += summary.totalExitedValidators;
+        }
+
+        if (_preCLBalance > _unifiedPostCLBalance) {
+            _addReportData(reportTimestamp, stakingRouterExitedValidators, _preCLBalance - _unifiedPostCLBalance);
+        } else {
+            _addReportData(reportTimestamp, stakingRouterExitedValidators, 0);
+            // If the CL balance is not decreased, we don't need to check anyting here
+            return;
+        }
```
TheDZhon, line 690: nitpick:

```suggestion
        if (_preCLBalance <= _unifiedPostCLBalance) {
             _addReportData(reportTimestamp, stakingRouterExitedValidators, 0);
            // If the CL balance is not decreased, we don't need to check anyting here
            return;
        }
        _addReportData(reportTimestamp, stakingRouterExitedValidators, _preCLBalance - _unifiedPostCLBalance);
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +633,96 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+        uint256 negativeCLRebaseSum = _sumNegativeRebasesNotOlderThan(reportTimestamp - 18 days);
+        uint256 maxCLRebaseNegativeSum =
```
TheDZhon, line 693:
```suggestion
        uint256 maxAllowedCLRebaseNegativeSum =
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +633,96 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+            if (clBalanceWei < _unifiedPostCLBalance) {
+                revert NegativeRebaseFailedCLBalanceMismatch(_unifiedPostCLBalance, clBalanceWei, _limitsList.clBalanceOraclesErrorUpperBPLimit);
+            }
+            if (MAX_BASIS_POINTS * (clBalanceWei - _unifiedPostCLBalance) >
+                _limitsList.clBalanceOraclesErrorUpperBPLimit * clBalanceWei) {
+                revert NegativeRebaseFailedCLBalanceMismatch(_unifiedPostCLBalance, clBalanceWei, _limitsList.clBalanceOraclesErrorUpperBPLimit);
+            }
+            emit NegativeCLRebaseConfirmed(_refSlot, _unifiedPostCLBalance);
+        } else {
+            revert NegativeRebaseFailedCLStateReportIsNotReady();
```
TheDZhon, line 725:
```suggestion
            revert NegativeRebaseFailedSecondOpinionReportIsNotReady();
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -715,13 +859,25 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
             _checkLimitValue(_newLimitsList.maxPositiveTokenRebase, 1, type(uint64).max);
             emit MaxPositiveTokenRebaseSet(_newLimitsList.maxPositiveTokenRebase);
         }
+        if (_oldLimitsList.initialSlashingAmountPWei != _newLimitsList.initialSlashingAmountPWei) {
+            _checkLimitValue(_newLimitsList.initialSlashingAmountPWei, 0, type(uint16).max);
+            emit InitialSlashingAmountSet(_newLimitsList.initialSlashingAmountPWei);
+        }
+        if (_oldLimitsList.inactivityPenaltiesAmountPWei != _newLimitsList.inactivityPenaltiesAmountPWei) {
+            _checkLimitValue(_newLimitsList.inactivityPenaltiesAmountPWei, 0, type(uint16).max);
+            emit PenaltiesAmountSet(_newLimitsList.inactivityPenaltiesAmountPWei);
```
TheDZhon, line 875:
```suggestion
            emit InactivityPenaltiesAmountSet(_newLimitsList.inactivityPenaltiesAmountPWei);
```

TheDZhon (approved):
Godspeed :rocket: 

Left some _minorish_ things only

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +636,97 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+        // If there is no negative rebase oracle, then we don't need to check it's report
+        if (address(secondOpinionOracle) == address(0)) {
+            // If there is no oracle and the diff is more than limit, we revert
+            revert IncorrectCLBalanceDecrease(negativeCLRebaseSum, maxAllowedCLRebaseNegativeSum);
+        }
+        _askSecondOpinion(_refSlot, _postCLBalance, _limitsList);
+    }
+
+    function _askSecondOpinion(uint256 _refSlot, uint256 _postCLBalance, LimitsList memory _limitsList) internal {
+        (bool success, uint256 clOracleBalanceGwei,,) = secondOpinionOracle.getReport(_refSlot);
```
TheDZhon, line 716: should it have the withdrawal vault balance included?

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -415,8 +487,12 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         uint256 _sharesRequestedToBurn,
         uint256 _preCLValidators,
         uint256 _postCLValidators
-    ) external view {
+    ) external {
+        if (msg.sender != LIDO_LOCATOR.lido()) {
```
TheDZhon, line 491: let's save LIDO in immutable var

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -45,7 +45,13 @@ interface ISecondOpinionOracle {
     function getReport(uint256 refSlot)
         external
         view
-        returns (bool success, uint256 clBalanceGwei, uint256 totalDepositedValidators, uint256 totalExitedValidators);
+        returns (
+            bool success,
+            uint256 clBalanceGwei,
+            uint256 withdrawalVaultBalance,
```
TheDZhon, line 51:
```suggestion
            uint256 withdrawalVaultBalanceWei,
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -675,7 +681,6 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         uint256 _postCLValidators,
         uint256 _refSlot
     ) internal {
-        uint256 unifiedPostCLBalance = _postCLBalance + _withdrawalVaultBalance;
```
TheDZhon: I am not sure if we need to remove this var and recalculate it here and there

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -711,11 +716,11 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
             revert IncorrectCLBalanceDecrease(negativeCLRebaseSum, maxAllowedCLRebaseNegativeSum);
         }
-        _askSecondOpinion(_refSlot, _postCLBalance, _limitsList);
+        _askSecondOpinion(_refSlot, _postCLBalance, _withdrawalVaultBalance, _limitsList);
     }
 
-    function _askSecondOpinion(uint256 _refSlot, uint256 _postCLBalance, LimitsList memory _limitsList) internal {
-        (bool success, uint256 clOracleBalanceGwei,,) = secondOpinionOracle.getReport(_refSlot);
+    function _askSecondOpinion(uint256 _refSlot, uint256 _postCLBalance, uint256 _withdrawalVaultBalance, LimitsList memory _limitsList) internal {
+        (bool success, uint256 clOracleBalanceGwei, uint256 oracleWithdrawalVaultBalance,,) = secondOpinionOracle.getReport(_refSlot);
```
TheDZhon, line 723: let's name it `oracleWithdrawalVaultBalanceWei`

TheDZhon (changes requested):
:eyes:

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: it's Solidity limitation on number of local vars

TheDZhon: I see

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +644,99 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+            IStakingRouter.StakingModuleSummary memory summary = stakingRouter.getStakingModuleSummary(ids[i]);
+            stakingRouterExitedValidators += summary.totalExitedValidators;
+        }
+
+        if (_preCLBalance <= _postCLBalance + _withdrawalVaultBalance) {
+            _addReportData(reportTimestamp, stakingRouterExitedValidators, 0);
+            // If the CL balance is not decreased, we don't need to check anyting here
+            return;
+        }
+        _addReportData(reportTimestamp, stakingRouterExitedValidators, _preCLBalance - _postCLBalance - _withdrawalVaultBalance);
```
TheDZhon, line 701:
```suggestion
        _addReportData(reportTimestamp, stakingRouterExitedValidators, _preCLBalance - (_postCLBalance + _withdrawalVaultBalance));
```

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +644,99 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+    ) internal {
+        uint256 reportTimestamp = GENESIS_TIME + _refSlot * SECONDS_PER_SLOT;
+
+        // Checking exitedValidators against StakingRouter
+        IStakingRouter stakingRouter = IStakingRouter(LIDO_LOCATOR.stakingRouter());
+        uint256[] memory ids = stakingRouter.getStakingModuleIds();
+
+        uint256 stakingRouterExitedValidators;
+        for (uint256 i = 0; i < ids.length; i++) {
+            IStakingRouter.StakingModuleSummary memory summary = stakingRouter.getStakingModuleSummary(ids[i]);
```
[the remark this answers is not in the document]
avsetsin: It would be great to have integration tests for checks related to exited validators since there's a pretty complex logic behind the delivery of that data.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -715,13 +873,25 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
         if (_oldLimitsList.requestTimestampMargin != _newLimitsList.requestTimestampMargin) {
-            _checkLimitValue(_newLimitsList.requestTimestampMargin, 0, type(uint64).max);
+            _checkLimitValue(_newLimitsList.requestTimestampMargin, 0, type(uint48).max);
             emit RequestTimestampMarginSet(_newLimitsList.requestTimestampMargin);
         }
         if (_oldLimitsList.maxPositiveTokenRebase != _newLimitsList.maxPositiveTokenRebase) {
             _checkLimitValue(_newLimitsList.maxPositiveTokenRebase, 1, type(uint64).max);
             emit MaxPositiveTokenRebaseSet(_newLimitsList.maxPositiveTokenRebase);
         }
+        if (_oldLimitsList.initialSlashingAmountPWei != _newLimitsList.initialSlashingAmountPWei) {
```
avsetsin, line 883: Nitpicking, feel free to ignore it. Consider spelling `Gwei` and `PWei` the same way (`PWei` -> `Pwei`)

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +644,99 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
     }
 
-    function _checkOneOffCLBalanceDecrease(
+    function _addReportData(uint256 _timestamp, uint256 _exitedValidatorsCount, uint256 _negativeCLRebase) internal {
+        reportData.push(ReportData(
+            SafeCast.toUint64(_timestamp),
+            SafeCast.toUint64(_exitedValidatorsCount),
+            SafeCast.toUint128(_negativeCLRebase)
+        ));
+    }
```
avsetsin, line 653: Since it mutates store consider emitting an event here

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +644,99 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+
+    function _sumNegativeRebasesNotOlderThan(uint256 _timestamp) internal view returns (uint256) {
+        uint256 sum;
+        for (int256 index = int256(reportData.length) - 1; index >= 0; index--) {
+            if (reportData[uint256(index)].timestamp > SafeCast.toUint64(_timestamp)) {
```
avsetsin, line 658: A small suggestion: the code can be simplified on int <> uint conversion if we put an element in `reportData` in the constructor beforehand. 

It safe to put there something like `(GENESIS_TIME, 0, 0)`

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -121,58 +178,52 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
     /// @param _lidoLocator address of the LidoLocator instance
     /// @param _admin address to grant DEFAULT_ADMIN_ROLE of the AccessControl contract
     /// @param _limitsList initial values to be set for the limits list
-    /// @param _managersRoster list of the address to grant permissions for granular limits management
     constructor(
         address _lidoLocator,
         address _admin,
-        LimitsList memory _limitsList,
-        ManagersRoster memory _managersRoster
+        LimitsList memory _limitsList
```
avsetsin, line 209: Changes in the constructor (new values in the `_limitsList` struct and getting rid of `_managersRoster`) should affect the scratch deployment, which is probably broken now

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -44,11 +85,6 @@ struct LimitsList {
     /// @dev Must fit into uint16 (<= 65_535)
     uint256 churnValidatorsPerDayLimit;
 
-    /// @notice The max decrease of the total validators' balances on the Consensus Layer since
-    ///     the previous oracle report
-    /// @dev Represented in the Basis Points (100% == 10_000)
-    uint256 oneOffCLBalanceDecreaseBPLimit;
```
avsetsin, line 50: Removing an element from the middle of the struct breaks a backward compatibility in the `getOracleReportLimits` method.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: That's for sure but it's intended to remove unused complexity in roles management

avsetsin: I mean, pls check scratch deployment scripts

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -415,8 +495,12 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
         uint256 _sharesRequestedToBurn,
         uint256 _preCLValidators,
         uint256 _postCLValidators
-    ) external view {
+    ) external {
+        if (msg.sender != LIDO_ADDRESS) {
+            revert CalledNotFromLido();
+        }
         LimitsList memory limitsList = _limits.unpack();
+        uint256 refSlot = IBaseOracle(LIDO_LOCATOR.accountingOracle()).getLastProcessingRefSlot();
```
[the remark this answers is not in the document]
vp4242: That's true. I would prefer the dependency could be avoided but that requires Lido contract change and formal re-audit for it.

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: Indeed deployment scripts requires changes. Will be fixed in a separate PR

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: Added comment about this to the spec https://github.com/lidofinance/lido-improvement-proposals/pull/61

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
vp4242: Actually applying deployment script fixes now

test/0.4.24/contracts/AccountingOracle__MockForLegacyOracle.sol
```diff
@@ -49,6 +49,10 @@ contract AccountingOracle__MockForLegacyOracle {
         );
     }
```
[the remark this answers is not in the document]
vp4242: It's no longer necessary as tests for negative rebase are using AccountingOracle__MockForSanityChecker

test/0.8.9/contracts/OracleReportSanityCheckerWrapper.sol
```diff
@@ -0,0 +1,42 @@
...
+// NB: for testing purposes only
+pragma solidity 0.8.9;
+
+import { OracleReportSanityChecker, LimitsList, LimitsListPacked, LimitsListPacker } from "../../../contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol";
+
+contract OracleReportSanityCheckerWrapper is OracleReportSanityChecker {
```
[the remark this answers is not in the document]
vp4242: "Harness" in naming it's a pretty recent naming practice. I would do it separately

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -77,26 +99,49 @@ struct LimitsList {
...
     /// @dev uses 1e9 precision, e.g.: 1e6 - 0.1%; 1e9 - 100%, see `setMaxPositiveTokenRebase()`
     uint256 maxPositiveTokenRebase;
+
+    /// @notice Initial slashing amount per one validator to calculate initial slashing of the validators' balances on the Consensus Layer
+    /// @dev Represented in the PWei (1^15 Wei). Must fit into uint16 (<= 65_535)
+    uint256 initialSlashingAmountPWei;
+
+    /// @notice Inactivity penalties amount per one validator to calculate penalties of the validators' balances on the Consensus Layer
+    /// @dev Represented in the PWei (1^15 Wei). Must fit into uint16 (<= 65_535)
+    uint256 inactivityPenaltiesAmountPWei;
```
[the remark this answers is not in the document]
vp4242: I would argue that max is a bit redundant here. The same logic could be applied for naming of initialSlashingAmountPWei. Seems that the change doesn't bring a lot of clarity to already long name

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -559,16 +630,101 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
...
+            // If the CL balance is not decreased, we don't need to check anything here
+            return;
+        }
+        _addReportData(reportTimestamp, stakingRouterExitedValidators, _preCLBalance - (_postCLBalance + _withdrawalVaultBalance));
+
+        // NOTE. Values of 18 and 54 days are taken from spec. Check the details here
+        // https://github.com/lidofinance/lido-improvement-proposals/blob/develop/LIPS/lip-23.md
+        uint256 negativeCLRebaseSum = _sumNegativeRebasesNotOlderThan(reportTimestamp - 18 days);
+        uint256 maxAllowedCLRebaseNegativeSum =
+            _limitsList.initialSlashingAmountPWei * ONE_PWEI * (_postCLValidators - _exitedValidatorsAtTimestamp(reportTimestamp - 18 days)) +
```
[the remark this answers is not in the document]
vp4242: It's 2200 of gas reduction in case of a single loop AFAIK. Do you think it's worth it for the current state?

TheDZhon (approved):
LGTM

contracts/0.8.9/interfaces/ISecondOpinionOracle.sol
```diff
@@ -0,0 +1,28 @@
+// SPDX-FileCopyrightText: 2024 Lido <info@lido.fi>
+// SPDX-License-Identifier: GPL-3.0
+
+// See contracts/COMPILERS.md
+// solhint-disable-next-line
+pragma solidity >=0.4.24 <0.9.0;
```
TheDZhon, line 6: :warning: On condition of using the wide pragma condition, it makes much more sense putting this interface to the `contracts/common/interfaces`

TheDZhon (changes requested):
:eyes:

TheDZhon (approved):
:ship:

PR #75, opened 2024-04-17
Port of Sanity check tests

vp4242:
Port of Sanity check tests
https://github.com/lidofinance/lido-dao/blob/master/test/0.8.9/oracle-report-sanity-checker.test.js

### DIFF
--- a/lib/units.ts
+++ b/lib/units.ts
@@ -6,6 +6,4 @@ const shares = (value: bigint) => parseUnits(value.toString(), "ether");
 
 const shareRate = (value: bigint) => parseUnits(value.toString(), 27);
 
-const ETH = (value: number) => ether(value.toString());
-
-export { ether, shares, shareRate, ETH };
+export { ether, shares, shareRate };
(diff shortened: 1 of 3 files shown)


lib/units.ts
```diff
@@ -6,4 +6,6 @@ const shares = (value: bigint) => parseUnits(value.toString(), "ether");
 
 const shareRate = (value: bigint) => parseUnits(value.toString(), 27);
 
-export { ether, shares, shareRate };
+const ETH = (value: number) => ether(value.toString());
```
mymphe, line 9: I strongly suspect that the `ETH` function will fail when working with large numbers. E.g.
```js
const a = 1000 * 10 ** 18;
a.toString() // "1e+21" not a valid input to `parseEther`
```

mymphe: I suggest sticking to `ether` and using string as input, e.g. `ether("1000.0")`

mymphe (changes requested):

mymphe (approved):

PR #78, opened 2024-04-18
Staking Router 1.5

avsetsin:


contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -326,92 +378,132 @@ contract DepositSecurityModule {
...
             guardianIndex = _getGuardianIndex(guardianAddr);
             if (guardianIndex == -1) revert InvalidSignature();
         }
 
-        if (block.number - blockNumber >  pauseIntentValidityPeriodBlocks) revert PauseIntentExpired();
+        if (block.number - blockNumber > pauseIntentValidityPeriodBlocks) revert PauseIntentExpired();
 
-        STAKING_ROUTER.pauseStakingModule(stakingModuleId);
-        emit DepositsPaused(guardianAddr, uint24(stakingModuleId));
+        isDepositsPaused = true;
```
F4ever, line 414: `depositBufferedEther` checks pauses of certain modules, but doesn't check `isDepositsPaused`. In this case malitious guardians can deposit again.

`depositBufferedEther` must call `canDeposit` or check isDepositsPaused
Or `pauseDeposits` should change all staking modules statuses to `DepositsPaused`

contracts/0.8.9/interfaces/IStakingModule.sol
```diff
@@ -23,7 +23,7 @@ interface IStakingModule {
 
     /// @notice Returns all-validators summary belonging to the node operator with the given id
     /// @param _nodeOperatorId id of the operator to return report for
-    /// @return isTargetLimitActive shows whether the current target limit applied to the node operator
+    /// @return targetLimitMode shows whether the current target limit applied to the node operator (1 = soft mode, 2 = forced mode)
```
dgusakov, line 26:
```suggestion
    /// @return targetLimitMode shows whether the current target limit applied to the node operator (0 = disabled, 1 = soft mode, 2 = forced mode)
```

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -60,44 +63,61 @@ contract DepositSecurityModule {
...
+    error ModuleNonceChanged();
     error PauseIntentExpired();
+    error UnvetIntentExpired();
+    error UnvetPayloadInvalid();
+    error UnvetUnexpectedBlockHash();
     error NotAGuardian(address addr);
     error ZeroParameter(string parameter);
 
+    /// @notice Represents the code version to help distinguish contract interfaces.
+    uint256 public constant VERSION = 3;
```
F4ever, line 76: Why version number is 3 if this is second version of dsm contract?

F4ever: Maybe add `getContractVersion` similar to contracts that inherits `Versioned` interface?

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -93,8 +93,8 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
     uint8 internal constant TOTAL_DEPOSITED_KEYS_COUNT_OFFSET = 3;
 
     // TargetValidatorsStats
-    /// @dev Flag enable/disable limiting target active validators count for operator
-    uint8 internal constant IS_TARGET_LIMIT_ACTIVE_OFFSET = 0;
+    /// @dev Target limit mode, allows limiting target active validators count for operator, >1 means forced target limit enabled
```
F4ever, line 96:
```suggestion
    /// @dev Target limit mode, allows limiting active validators count for operator, >1 means forced target limit enabled
```
Maybe explicitly indicate 2 for forced target limit, in case if we use new modes there will be no compatibility issues

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -580,18 +631,21 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
     /// @notice Updates the limit of the validators that can be used for deposit by DAO
     /// @param _nodeOperatorId Id of the node operator
     /// @param _targetLimit Target limit of the node operator
-    /// @param _isTargetLimitActive active flag
-    function updateTargetValidatorsLimits(uint256 _nodeOperatorId, bool _isTargetLimitActive, uint256 _targetLimit) external {
+    /// @param _targetLimitMode target limit mode, >1 means forced target limit
+    function updateTargetValidatorsLimits(uint256 _nodeOperatorId, uint256 _targetLimitMode, uint256 _targetLimit) public {
```
F4ever, line 635: Why changed to public?

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -580,18 +631,21 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
...
         _requireValidRange(_targetLimit <= UINT64_MAX);
 
         Packed64x4.Packed memory operatorTargetStats = _loadOperatorTargetValidatorsStats(_nodeOperatorId);
-        operatorTargetStats.set(IS_TARGET_LIMIT_ACTIVE_OFFSET, _isTargetLimitActive ? 1 : 0);
-        operatorTargetStats.set(TARGET_VALIDATORS_COUNT_OFFSET, _isTargetLimitActive ? _targetLimit : 0);
+        operatorTargetStats.set(TARGET_LIMIT_MODE_OFFSET, _targetLimitMode);
+        if (_targetLimitMode == 0) {
+            _targetLimit = 0;
+        }
+        operatorTargetStats.set(TARGET_VALIDATORS_COUNT_OFFSET, _targetLimit);
```
F4ever, line 645: Why not use previous version?
`operatorTargetStats.set(TARGET_VALIDATORS_COUNT_OFFSET, _targetLimitMode ? _targetLimit : 0);`

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -121,14 +141,25 @@ contract DepositSecurityModule {
...
+                address(this)
+            )
+        );
+
         _setOwner(msg.sender);
-        _setMaxDeposits(_maxDepositsPerBlock);
-        _setMinDepositBlockDistance(_minDepositBlockDistance);
+        _setLastDepositBlock(block.number);
         _setPauseIntentValidityPeriodBlocks(_pauseIntentValidityPeriodBlocks);
+        _setUnvetIntentValidityPeriodBlocks(_unvetIntentValidityPeriodBlocks);
```
F4ever, line 156: Is there a case when this period is useful?
Pause validity period is set to prevent new pause, but unvet messages can be not relevant only because of staking module nonce change.

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -460,11 +554,91 @@ contract DepositSecurityModule {
...
+                    nodeOperatorIds,
+                    vettedSigningKeysCounts
+                )
+            );
+            guardianAddr = ECDSA.recover(msgHash, sig.r, sig.vs);
+            guardianIndex = _getGuardianIndex(guardianAddr);
+            if (guardianIndex == -1) revert InvalidSignature();
+        }
+
+        if (blockHash == bytes32(0) || blockhash(blockNumber) != blockHash) revert UnvetUnexpectedBlockHash();
```
F4ever, line 635: What is the reason to check `blockHash == bytes32(0)`?

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -580,18 +631,21 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
...
+    /// @param _targetLimitMode target limit mode, >1 means forced target limit
+    function updateTargetValidatorsLimits(uint256 _nodeOperatorId, uint256 _targetLimitMode, uint256 _targetLimit) public {
         _onlyExistedNodeOperator(_nodeOperatorId);
         _auth(STAKING_ROUTER_ROLE);
         _requireValidRange(_targetLimit <= UINT64_MAX);
 
         Packed64x4.Packed memory operatorTargetStats = _loadOperatorTargetValidatorsStats(_nodeOperatorId);
-        operatorTargetStats.set(IS_TARGET_LIMIT_ACTIVE_OFFSET, _isTargetLimitActive ? 1 : 0);
-        operatorTargetStats.set(TARGET_VALIDATORS_COUNT_OFFSET, _isTargetLimitActive ? _targetLimit : 0);
+        operatorTargetStats.set(TARGET_LIMIT_MODE_OFFSET, _targetLimitMode);
```
F4ever, line 641: Maybe validate `_targetLimitMode` is in [0, 1, 2]

contracts/0.8.9/StakingRouter.sol
```diff
@@ -16,1305 +16,1351 @@ import {BeaconChainDepositor} from "./BeaconChainDepositor.sol";
...
+    for (uint256 i = 0; i < _stakingModuleIds.length; ) {
+      uint256 stakingModuleId = _stakingModuleIds[i];
+      StakingModule storage stakingModule = _getStakingModuleById(stakingModuleId);
+
+      uint256 prevReportedExitedValidatorsCount = stakingModule.exitedValidatorsCount;
+      if (_exitedValidatorsCounts[i] < prevReportedExitedValidatorsCount) {
+        revert ExitedValidatorsCountCannotDecrease();
+      }
+
+      (uint256 totalExitedValidators, uint256 totalDepositedValidators, ) = /* uint256 depositableValidatorsCount */
```
F4ever, line 482: Previous version was better 
```
(
    uint256 totalExitedValidators, 
    uint256 totalDepositedValidators,
/* uint256 depositableValidatorsCount */
) = IStakingModule(stakingModule.stakingModuleAddress).getStakingModuleSummary();
```

F4ever: Or just remove comment

contracts/0.8.9/DepositSecurityModule.sol
avsetsin: fixed

contracts/0.8.9/DepositSecurityModule.sol
avsetsin: Removed unvet intent period. The messages will be invalidated in 256 blocks since there is a limit to retrieve blockhash of historical blocks. So the limit will be the same as it for attestation messages.

contracts/0.8.9/DepositSecurityModule.sol
avsetsin: There is a limit to retrieve blockhash for historical blocks. If you try to retrieve blockhash for a block older than now - 256 blocks you will get zero bytes. So to be sure that blockhash is valid and matches the onchain one we need to check that it's not equal to zero bytes

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
krogla: no need for a while, as it just like a flag `forced=mode>1` for VEBO

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
krogla: due to `external` method not allows to modify input parameter values even if they are simple types (e.g. `uint256`).  So it's kind of a "visibility+9 bytes of code  V/S  new var + some extra gas per call" trade-off.

tamtamchik:
JFYI, foundry tests fail because of some s*ty foundry nightly build came to CI; everything works on later versions (just checked with @eddort). Hopefully will fix itself tomorrow... 😢

tamtamchik:
I dived deeper into foundry error because I couldn't reproduce it in `develop`, and this commit looks like is a major setback for the forge tests: https://github.com/lidofinance/core/pull/78/commits/2e783bd8286da84a41daa0ae20167a28ac46cac2

STR:
- Update foundry to somewhat latest >2024-06-03
- Check out this commit 
- Do: `forge compile` and then `forge test` – get an error
- Reverting the changes in the commit
- Do: `forge compile` and then `forge test` – no error 

I'm pretty sure the problem is somewhere on the foundry side, just before creating an issue for them want to confirm that I'm not missing something...

cc @avsetsin @krogla

mkurayan:
> I dived deeper into foundry error because I couldn't reproduce it in `develop`, and this commit looks like is a major setback for the forge tests: [2e783bd](https://github.com/lidofinance/core/commit/2e783bd8286da84a41daa0ae20167a28ac46cac2)
> 
> STR:
> 
> * Update foundry to somewhat latest >2024-06-03
> * Check out this commit
> * Do: `forge compile` and then `forge test` – get an error
> * Reverting the changes in the commit
> * Do: `forge compile` and then `forge test` – no error
> 
> I'm pretty sure the problem is somewhere on the foundry side, just before creating an issue for them want to confirm that I'm not missing something...
> 
> cc @avsetsin @krogla

I re-run job, now all tests passed, looks like it was some foundry issue indeed.

test/common/minFirstAllocationStrategy.t.sol
AlexandrMov, line 1: test?

contracts/0.8.9/StakingRouter.sol
```diff
@@ -299,8 +407,7 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
 
         for (uint256 i = 0; i < _stakingModuleIds.length; ) {
             if (_totalShares[i] > 0) {
-                address moduleAddr = _getStakingModuleById(_stakingModuleIds[i]).stakingModuleAddress;
-                try IStakingModule(moduleAddr).onRewardsMinted(_totalShares[i]) {}
+                try _getIStakingModuleById(_stakingModuleIds[i]).onRewardsMinted(_totalShares[i]) {}
                 catch (bytes memory lowLevelRevertData) {
```
dgusakov, line 411: With the upgraded version of NoR this might be redundant

contracts/0.8.9/StakingRouter.sol
```diff
@@ -485,41 +588,41 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
...
         onlyRole(UNSAFE_SET_EXITED_VALIDATORS_ROLE)
     {
-        StakingModule storage stakingModule = _getStakingModuleById(_stakingModuleId);
-        address moduleAddr = stakingModule.stakingModuleAddress;
+        StakingModule storage stakingModuleState = _getStakingModuleById(_stakingModuleId);
+        IStakingModule stakingModule = IStakingModule(stakingModuleState.stakingModuleAddress);
 
         (
-            /* bool isTargetLimitActive */,
+            /* uint156 targetLimitMode */,
```
dgusakov, line 595: Why 156?

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -133,6 +133,8 @@ contract AccountingOracle is BaseOracle {
     bytes32 internal constant EXTRA_DATA_PROCESSING_STATE_POSITION =
         keccak256("lido.AccountingOracle.extraDataProcessingState");
 
+    bytes32 internal constant ZERO_HASH = bytes32(0);
```
dgusakov, line 136: I'm not sure, but it looks like the compiler can handle it without explicit consts. At least with 0.8.24

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -747,9 +771,9 @@ contract OracleReportSanityChecker is AccessControlEnumerable {
     error IncorrectSharesRequestedToBurn(uint256 actualSharesToBurn);
     error IncorrectCLBalanceDecrease(uint256 oneOffCLBalanceDecreaseBP);
     error IncorrectCLBalanceIncrease(uint256 annualBalanceDiff);
-    error IncorrectAppearedValidators(uint256 churnLimit);
+    error IncorrectAppearedValidators(uint256 appearedValidatorsLimit);
     error IncorrectNumberOfExitRequestsPerReport(uint256 maxRequestsCount);
-    error IncorrectExitedValidators(uint256 churnLimit);
+    error IncorrectExitedValidators(uint256 exitedValudatorsLimit);
```
skhomuti, line 776:
```suggestion
    error IncorrectExitedValidators(uint256 exitedValidatorsLimit);
```

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -280,6 +295,17 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
         emit StakingModuleTypeSet(_type);
     }
 
+    function finalizeUpgrade_v3() external {
```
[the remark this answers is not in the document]
mkurayan: I asked the same question, if i correctly remember @krogla  suggested to keep it, because it a migration history. 
So you have clear path how to migrate v1 -> v2 ->v3.

We can remove if you think it not useful.

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -390,6 +416,55 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
         _authP(SET_NODE_OPERATOR_LIMIT_ROLE, arr(uint256(_nodeOperatorId), uint256(_vettedSigningKeysCount)));
         _onlyCorrectNodeOperatorState(getNodeOperatorIsActive(_nodeOperatorId));
 
+        _updateVettedSingingKeysCount(_nodeOperatorId, _vettedSigningKeysCount, true /* _allowIncrease */);
+        _increaseValidatorsKeysNonce();
```
AlexandrMov, line 420: I think tooling services don't use `keysop` index anymore?

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -390,6 +416,55 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
         _authP(SET_NODE_OPERATOR_LIMIT_ROLE, arr(uint256(_nodeOperatorId), uint256(_vettedSigningKeysCount)));
```
AlexandrMov, line 416: probably, make function name more clear?

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -42,13 +44,13 @@ contract DepositSecurityModule {
 
     event OwnerChanged(address newValue);
     event PauseIntentValidityPeriodBlocksChanged(uint256 newValue);
-    event MaxDepositsChanged(uint256 newValue);
-    event MinDepositBlockDistanceChanged(uint256 newValue);
+    event MaxOperatorsPerUnvettingChanged(uint256 newValue);
     event GuardianQuorumChanged(uint256 newValue);
     event GuardianAdded(address guardian);
     event GuardianRemoved(address guardian);
-    event DepositsPaused(address indexed guardian, uint24 indexed stakingModuleId);
```
AlexandrMov, line 50: is there any way to figure out `stakingModuleId` to format alert message accordingly?

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -852,7 +963,7 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
             }
         }
 
-        allocatedKeysCount =
+        (allocatedKeysCount, activeKeyCountsAfterAllocation) =
```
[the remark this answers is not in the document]
krogla: no, this is a effect due to the MinFirstAllocationStrategy library becoming external and memory data structures are passed by copying rather than by reference

contracts/0.8.9/StakingRouter.sol
krogla: typo)

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -580,18 +688,21 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
     /// @notice Updates the limit of the validators that can be used for deposit by DAO
     /// @param _nodeOperatorId Id of the node operator
     /// @param _targetLimit Target limit of the node operator
-    /// @param _isTargetLimitActive active flag
-    function updateTargetValidatorsLimits(uint256 _nodeOperatorId, bool _isTargetLimitActive, uint256 _targetLimit) external {
+    /// @param _targetLimitMode target limit mode (0 = disabled, 1 = soft mode, 2 = forced mode)
+    function updateTargetValidatorsLimits(uint256 _nodeOperatorId, uint256 _targetLimitMode, uint256 _targetLimit) public {
         _onlyExistedNodeOperator(_nodeOperatorId);
         _auth(STAKING_ROUTER_ROLE);
         _requireValidRange(_targetLimit <= UINT64_MAX);
```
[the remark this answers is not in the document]
krogla: it seems that it is not necessary, since this parameter is needed only for offchain tooling, and there is no agreement anywhere in the onchain code that it should be limited, and the value is only checked for inequality 0, so there is no sense in doing a hardcoded check

contracts/0.8.9/StakingRouter.sol
```diff
@@ -228,67 +290,115 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
         STAKING_MODULES_COUNT_POSITION.setStorageUint256(newStakingModuleIndex + 1);
 
         emit StakingModuleAdded(newStakingModuleId, _stakingModuleAddress, _name, msg.sender);
-        emit StakingModuleTargetShareSet(newStakingModuleId, _targetShare, msg.sender);
-        emit StakingModuleFeesSet(newStakingModuleId, _stakingModuleFee, _treasuryFee, msg.sender);
+        _updateStakingModule(
```
AlexandrMov, line 293: does it make sense to make an update first rather than emitting?

contracts/0.8.9/StakingRouter.sol
```diff
@@ -726,7 +846,7 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
         /// @dev using intermediate variables below due to "Stack too deep" error in case of
         ///     assigning directly into the NodeOperatorSummary struct
         (
-            bool isTargetLimitActive,
+            uint256 targetLimitMode,
```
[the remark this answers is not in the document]
krogla: yes it is, but this change is more required for external tooling, and afaik, no onchain code uses this value directly. 
The behavior and backwards compatibility of offchain tooling has already been tested

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
krogla: it's a legacy part of the Aragon framework so we can't do anything about

contracts/0.8.9/StakingRouter.sol
krogla: within a single transaction is a matter of taste, but I find the "first do, then report" flow more logical )

contracts/0.8.9/StakingRouter.sol
```diff
@@ -582,41 +688,58 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
...
+        uint256 _stakingModuleId,
+        bytes calldata _nodeOperatorIds,
+        bytes calldata _vettedSigningKeysCounts
+    ) external onlyRole(STAKING_MODULE_UNVETTING_ROLE) {
+        _checkValidatorsByNodeOperatorReportData(_nodeOperatorIds, _vettedSigningKeysCounts);
+        _getIStakingModuleById(_stakingModuleId).decreaseVettedSigningKeysCount(_nodeOperatorIds, _vettedSigningKeysCounts);
+    }
+
+    /// @notice Returns all registered staking modules.
+    /// @return res Array of staking modules.
```
AlexandrMov, line 713:
```suggestion
    /// @return Array of staking modules.
```

contracts/0.8.9/StakingRouter.sol
```diff
@@ -1051,45 +1195,48 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
         }
     }
 
-    /// @notice Helper for Lido contract (DEPRECATED)
-    ///         Returns total fee total fee to mint for each staking
-    ///         module and treasury in reduced, 1e4 precision.
-    ///         In integrations please use getStakingRewardsDistribution().
-    ///         reduced, 1e4 precision.
+    /// @notice Returns the same as getStakingRewardsDistribution() but in reduced, 1e4 precision (DEPRECATED).
```
AlexandrMov, line 1198: is it not right time to remove DEPRECATED method?

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -711,56 +729,75 @@ contract AccountingOracle is BaseOracle {
...
         _checkCanSubmitExtraData(procState, EXTRA_DATA_FORMAT_LIST);
 
         if (procState.itemsProcessed == procState.itemsCount) {
             revert ExtraDataAlreadyProcessed();
         }
 
-        if (procState.itemsProcessed != 0) {
-            revert ExtraDataListOnlySupportsSingleTx();
+        // at least 32 bytes for the next hash value + 35 bytes for the first item with 1 node operator
+        if(data.length < 67) {
```
AlexandrMov, line 741:
```suggestion
        if (data.length < 67) {
```

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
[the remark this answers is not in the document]
krogla: looks like we don't plan to repeat upgrade v1->v2, and afaik there is no protocol v1 fork in a wild, so lets remove it, or just comment out it to keep historical code

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -399,25 +474,25 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
             totalSigningKeysCount, Math256.max(_vettedSigningKeysCount, depositedSigningKeysCount)
         );
 
-        if (vettedSigningKeysCountAfter == vettedSigningKeysCountBefore) {
-            return;
-        }
+        if (vettedSigningKeysCountAfter == vettedSigningKeysCountBefore) return;
```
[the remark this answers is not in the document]
krogla: this condition is mainly intended to skip unnecessary storage updates during the loop,
but obviously guardians should not report the same vetted count in general

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -511,6 +586,41 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
...
+    /// 1. Oracle first phase: Reach hash consensus.
+    /// 2. Oracle second phase: Module receives rewards.
+    ///
+    ///                                ... Reward CANNOT be distributed ...
+    ///
+    /// 3. Oracle third phase: Operator statistics are updated.
+    ///
+    ///                                 ... Reward can be distributed ...
+    ///
+    /// =====================================  Start report frame 3  =====================================
```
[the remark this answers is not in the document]
mkurayan: Sure, the method description has been updated. Let me know if you want any further changes, and I will update it accordingly.

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
krogla: DSM uses

contracts/0.8.9/oracle/AccountingOracle.sol
krogla: compiler requires constants initialization, even though it replaces them with an absolute value at compile time)
so it's just here for code readability

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -460,11 +528,92 @@ contract DepositSecurityModule {
...
-        for (uint256 i = 0; i < sigs.length; ++i) {
+        for (uint256 i = 0; i < sigs.length; ) {
             address signerAddr = ECDSA.recover(msgHash, sigs[i].r, sigs[i].vs);
             if (!_isGuardian(signerAddr)) revert InvalidSignature();
             if (signerAddr <= prevSignerAddr) revert SignaturesNotSorted();
             prevSignerAddr = signerAddr;
+
+            unchecked {
+                ++i;
+            }
```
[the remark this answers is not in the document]
krogla: moreover! since solc 0.8.22 unchecked loop increment optimizer are always enabled by default, so this technique does not give a gain in gas, however it gives -33 bytes in code size, but in the context of this small contract it seems unnecessary :)

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -326,92 +355,136 @@ contract DepositSecurityModule {
...
+     */
+    function isMinDepositDistancePassed(uint256 stakingModuleId) external view returns (bool) {
+        return _isMinDepositDistancePassed(stakingModuleId);
+    }
+
+    function _isMinDepositDistancePassed(uint256 stakingModuleId) internal view returns (bool) {
+        /// @dev The distance is reset when a deposit is made to any module. This prevents a front-run attack
+        /// by colluding guardians on several modules at once, providing the necessary window for an honest
+        /// guardian to react and pause deposits to all modules.
+        uint256 lastDepositToModuleBlock = STAKING_ROUTER.getStakingModuleLastDepositBlock(stakingModuleId);
```
krogla, line 463: non-optimal double call of SR reading the same storage slot

suggest a single getter in the SR contract that returns the `tuple(minDepositBlockDistance, maxDepositsPerBlock)`

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -326,92 +351,132 @@ contract DepositSecurityModule {
...
         bool isLidoCanDeposit = LIDO.canDeposit();
+
         return (
-            isModuleActive
+            !isDepositsPaused
+            && isModuleActive
             && quorum > 0
-            && block.number - lastDepositBlock >= minDepositBlockDistance
+            && isDepositDistancePassed
             && isLidoCanDeposit
```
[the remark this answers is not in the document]
krogla: since this method is not used on-chain, the order does not make much difference, besides, optimizing the load on the RPC node is beyond the scope of this PR :)

moreover, in case to get a `true` result condition, we have to read the state of all variables, so again the order does not matter

contracts/0.8.9/StakingRouter.sol
krogla: keep it for general compatibility

contracts/0.8.9/StakingRouter.sol
krogla: method still used in Lido contract which we don't want to update now

contracts/0.8.9/DepositSecurityModule.sol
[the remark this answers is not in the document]
avsetsin: Seems better to leave the current one. Most of the DSM code is well audited earlier with version 0.8.9 by different auditors

contracts/0.8.9/DepositSecurityModule.sol
avsetsin: Now it will be pause for the entire deposits to all modules. Note, that module status is remaning: it can be paused, stopped or activated but now only by DAO, not by DSM

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -390,6 +416,55 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
...
         _onlyCorrectNodeOperatorState(getNodeOperatorIsActive(_nodeOperatorId));
 
+        _updateVettedSingingKeysCount(_nodeOperatorId, _vettedSigningKeysCount, true /* _allowIncrease */);
+        _increaseValidatorsKeysNonce();
+    }
+
+    /// @notice Called by StakingRouter to decrease the number of vetted keys for node operator with given id
+    /// @param _nodeOperatorIds bytes packed array of the node operators id
+    /// @param _vettedSigningKeysCounts bytes packed array of the new number of vetted keys for the node operators
+    function decreaseVettedSigningKeysCount(
```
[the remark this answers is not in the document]
eddort: covered in https://github.com/lidofinance/core/pull/122

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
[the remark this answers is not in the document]
mkurayan: After a brief discussion, we decided to keep finalizeUpgrade_v2 for now. However, it may be deleted later if there isn't enough space in the NOR contract.

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -173,80 +205,66 @@ contract DepositSecurityModule {
...
-     *
-     * NB: the value must be harmonized with `OracleReportSanityChecker.churnValidatorsPerDayLimit`
-     * (see docs for the `OracleReportSanityChecker.setChurnValidatorsPerDayLimit` function)
+     * @notice Sets the maximum number of operators per unvetting.
+     * @param newValue The new maximum number of operators per unvetting.
+     * @dev Only callable by the owner.
      */
-    function setMaxDeposits(uint256 newValue) external onlyOwner {
-        _setMaxDeposits(newValue);
+    function setMaxOperatorsPerUnvetting(uint256 newValue) external onlyOwner {
```
ujenjt, line 220: I suggest to add note that this is really technical constraint to avoid unbound loop issue and meaningful value for this param would be ~200 or so...

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -458,13 +534,95 @@ contract DepositSecurityModule {
...
+        bytes calldata vettedSigningKeysCounts,
+        Signature calldata sig
+    ) external {
+        /// @dev The most likely reason for the signature to go stale
+        uint256 onchainNonce = STAKING_ROUTER.getStakingModuleNonce(stakingModuleId);
+        if (nonce != onchainNonce) revert ModuleNonceChanged();
+
+        uint256 nodeOperatorsCount = nodeOperatorIds.length / 8;
+
+        if (
```
ujenjt, line 588: it's unclear from first glance that the question is in the way how the line might be cut in pieces

contracts/0.8.9/StakingRouter.sol
```diff
@@ -1188,10 +1343,24 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
...
+    /// @dev Save the last deposit state for the staking module and emit the event
+    /// @param stakingModule staking module storage ref
+    /// @param stakingModuleId id of the staking module to be deposited
+    /// @param depositsValue value to deposit
+    function _updateModuleLastDepositState(
+        StakingModule storage stakingModule,
+        uint256 stakingModuleId,
+        uint256 depositsValue
+    ) internal {
+        stakingModule.lastDepositAt = uint64(block.timestamp);
```
ujenjt, line 1355: Do we need this counter cause we have this `_setLastDepositBlock(block.number);` within deposit security module.

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -258,14 +268,22 @@ contract AccountingOracle is BaseOracle {
...
         /// arrives.
         ///
-        /// Extra data is an array of items, each item being encoded as follows:
+        /// Depending on the size of the extra data, the processing might need to be split into
+        /// multiple transactions. Each transaction contains a chunk of report data (an array of items)
+        /// and the hash of the next transaction. The last transaction will contain ZERO_HASH
+        /// as the next transaction hash.
+        ///
+        /// | 32 bytes |    array of items
+        /// | nextHash |         ...
```
ujenjt, line 284: please update the comment within line 361, cause ` bytes32 extraDataHash;` is now hash of first item in the chain but not hash of the whole extra data

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -711,56 +728,70 @@ contract AccountingOracle is BaseOracle {
...
+        bytes32 dataHash = keccak256(data);
         if (dataHash != procState.dataHash) {
             revert UnexpectedExtraDataHash(procState.dataHash, dataHash);
         }
 
+        // load the next hash value
+        assembly {
+            dataHash := calldataload(data.offset)
+        }
+
```
ujenjt, line 748: why so hacky? 
```
bytes memory callData = msg.data; 
bytes  dataHash = callData[4:36];
```

ujenjt: If possible let's simplify it

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -772,7 +803,7 @@ contract AccountingOracle is BaseOracle {
                 dataOffset := add(dataOffset, 5)
             }
 
-            if (iter.itemType == 0) {
+            if (iter.lastSortingKey == 0) {
```
ujenjt, line 806: what? looks like it's overcomplicated logic, or even worse might be wrong for some edge cases

ujenjt: double checked - it should work but it's very ugly

ujenjt: if this is just checking the increase and sequence of item indices between transactions in the third phase, then please make it more semantic

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -711,56 +728,70 @@ contract AccountingOracle is BaseOracle {
...
         }
 
+        // load the next hash value
+        assembly {
+            dataHash := calldataload(data.offset)
+        }
+
         ExtraDataIterState memory iter = ExtraDataIterState({
-            index: 0,
+            index: procState.itemsProcessed > 0 ? procState.itemsProcessed - 1 : 0,
```
ujenjt, line 750: 🤔

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -258,14 +268,22 @@ contract AccountingOracle is BaseOracle {
...
         /// attached to some data report is processed in full before the processing deadline expires
         /// or a new data report starts being processed, but enforces that no processing of extra
         /// data for a report is possible after its processing deadline passes or a new data report
         /// arrives.
         ///
-        /// Extra data is an array of items, each item being encoded as follows:
+        /// Depending on the size of the extra data, the processing might need to be split into
+        /// multiple transactions. Each transaction contains a chunk of report data (an array of items)
+        /// and the hash of the next transaction. The last transaction will contain ZERO_HASH
+        /// as the next transaction hash.
```
skozin, line 281: i'd avoid using the term "transaction hash" since it means a specific and a different thing within the blockchain context. id suggest wording it like this:

> Each transaction contains a chunk of report data (an array of items) and the hash of the next chunk. The last transaction will contain ZERO_HASH as the next chunk hash.

contracts/0.8.9/oracle/AccountingOracle.sol
skozin: i'd simplify by renaming `iter.index` to `iter.nextIndex`, passing `procState.itemsProcessed` unconditionally here, and changing the check on lines 806-811 to `if (index != iter.nextIndex)`

contracts/0.8.9/oracle/AccountingOracle.sol
skozin: `if (index != iter.nextIndex)` (see my comment for line 750)

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -711,56 +728,70 @@ contract AccountingOracle is BaseOracle {
...
         uint256 dataOffset = iter.dataOffset;
         uint256 maxNodeOperatorsPerItem = 0;
         uint256 maxNodeOperatorItemIndex = 0;
+        uint256 itemsCount;
+        uint256 index;
+        uint256 itemType;
 
         while (dataOffset < data.length) {
-            uint256 index;
-            uint256 itemType;
```
skozin, line 762: what's the reason for moving these two variables to a broader scope?

contracts/0.8.9/StakingRouter.sol
```diff
@@ -67,83 +82,93 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
...
     bytes32 public constant STAKING_MODULE_MANAGE_ROLE = keccak256("STAKING_MODULE_MANAGE_ROLE");
+    bytes32 public constant STAKING_MODULE_UNVETTING_ROLE = keccak256("STAKING_MODULE_UNVETTING_ROLE");
     bytes32 public constant REPORT_EXITED_VALIDATORS_ROLE = keccak256("REPORT_EXITED_VALIDATORS_ROLE");
     bytes32 public constant UNSAFE_SET_EXITED_VALIDATORS_ROLE = keccak256("UNSAFE_SET_EXITED_VALIDATORS_ROLE");
     bytes32 public constant REPORT_REWARDS_MINTED_ROLE = keccak256("REPORT_REWARDS_MINTED_ROLE");
 
     bytes32 internal constant LIDO_POSITION = keccak256("lido.StakingRouter.lido");
 
-    /// @dev Credentials which allows the DAO to withdraw Ether on the 2.0 side
+    /// @dev Credentials to withdraw ETH on Consensus Layer side.
```
TheDZhon, line 141: :heart:

contracts/0.8.9/StakingRouter.sol
```diff
@@ -207,100 +270,140 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
...
         ///      https://docs.soliditylang.org/en/v0.8.17/types.html#enums
         newStakingModule.status = uint8(StakingModuleStatus.Active);
 
-        /// @dev  Simulate zero value deposit to prevent real deposits into the new StakingModule via
-        ///       DepositSecurityModule just after the addition.
-        ///       See DepositSecurityModule.getMaxDeposits() for details
-        newStakingModule.lastDepositAt = uint64(block.timestamp);
-        newStakingModule.lastDepositBlock = block.number;
-        emit StakingRouterETHDeposited(newStakingModuleId, 0);
+        /// @dev Simulate zero value deposit to prevent real deposits into the new StakingModule via
```
TheDZhon, line 279: Why it's called 'simulate'? maybe 'initialize' or 'pretend'?

TheDZhon: Why should it be prevented, though?

contracts/0.8.9/oracle/AccountingOracle.sol
```diff
@@ -578,14 +598,11 @@ contract AccountingOracle is BaseOracle {
...
-             if (data.extraDataHash == bytes32(0)) {
+            if (data.extraDataHash == ZERO_HASH) {
                 revert ExtraDataHashCannotBeZeroForNonEmptyData();
             }
         }
 
-        IOracleReportSanityChecker(LOCATOR.oracleReportSanityChecker())
-            .checkAccountingExtraDataListItemsCount(data.extraDataItemsCount);
-
         ILegacyOracle(LEGACY_ORACLE).handleConsensusLayerReport(
```
TheDZhon, line 606: It would have been great having LEGACY_ORACLE removed completely :cry: 
Let's schedule it for the next upgrade

contracts/0.8.9/sanity_checks/OracleReportSanityChecker.sol
```diff
@@ -105,13 +102,14 @@ struct LimitsList {
 
 /// @dev The packed version of the LimitsList struct to be effectively persisted in storage
 struct LimitsListPacked {
```
TheDZhon, line 104: cool, still fits 256bits :wine_glass:

contracts/common/lib/MinFirstAllocationStrategy.sol
```diff
@@ -31,7 +31,7 @@ library MinFirstAllocationStrategy {
         uint256[] memory buckets,
         uint256[] memory capacities,
         uint256 allocationSize
-    ) internal pure returns (uint256 allocated) {
+    ) public pure returns (uint256 allocated, uint256[] memory) {
```
TheDZhon, line 34: Please don't forget to add docs and artifacts bc now it's a separately deploying thing

TheDZhon (approved):
LGTM except only a few things for the future

contracts/common/lib/MinFirstAllocationStrategy.sol
avsetsin: Docs updated

contracts/0.8.9/DepositSecurityModule.sol
```diff
@@ -326,92 +355,136 @@ contract DepositSecurityModule {
...
+     * Reverts if deposits are not paused.
      */
-    function unpauseDeposits(uint256 stakingModuleId) external onlyOwner {
-         /// @dev unpause only paused modules (skip stopped)
-        if (STAKING_ROUTER.getStakingModuleIsDepositsPaused(stakingModuleId)) {
-            STAKING_ROUTER.resumeStakingModule(stakingModuleId);
-            emit DepositsUnpaused(uint24(stakingModuleId));
-        }
+    function unpauseDeposits() external onlyOwner {
+        if (!isDepositsPaused) revert DepositsNotPaused();
```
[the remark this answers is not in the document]
avsetsin: Yes, there is asymmetry in these methods. This behavior is done intentionally and the reason why early exit for pause is done is documented in the `pauseDeposits` method

https://github.com/lidofinance/core/blob/737b675c6a7c6c9190a619bd001d1a34a3f28368/contracts/0.8.9/DepositSecurityModule.sol#L372-L377

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -1254,6 +1355,9 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
         _saveOperatorStuckPenaltyStats(_nodeOperatorId, stuckPenaltyStats);
         _updateSummaryMaxValidatorsCount(_nodeOperatorId);
         _increaseValidatorsKeysNonce();
+
+        emit NodeOperatorPenaltyCleared(_nodeOperatorId);
+        return true;
```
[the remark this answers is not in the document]
avsetsin: This is a fix for a technical debt. The method interface assumed return bool, but always returned false because there was no return statement. This was found during a security audit. We preferred to add `return: true` and keep the method interface

contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -280,6 +295,22 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
         emit StakingModuleTypeSet(_type);
     }
 
+    function finalizeUpgrade_v3() external {
```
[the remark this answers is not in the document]
avsetsin: Good point, will put it in the backlog

contracts/0.8.9/StakingRouter.sol
avsetsin: To make sure that the minimum distance is kept between 2 sequential deposits in rare edge cases. For example, if we add a module and update DSM contract, as for example in SR+CSM update.

PR #84, opened 2024-04-23
chore: move to npm

tamtamchik:


tamtamchik:
https://github.com/npm/rfcs/issues/287 – is a showstopper since we have some mono repo dependencies in our setup. This leads to errors during installation and running.

> npm ERR! npm ERR! Error: Patch file found for package contract-helpers-test which is not present at node_modules/@aragon/contract-helpers-test

PR #111, opened 2024-06-19
sr1.5 - calls optimization

krogla:


contracts/0.8.9/StakingRouter.sol
```diff
@@ -1051,8 +1049,10 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
             StakingModuleCache[] memory stakingModulesCache
         ) = _getDepositsAllocation(_maxDepositsValue / DEPOSIT_SIZE);
         uint256 stakingModuleIndex = _getStakingModuleIndexById(_stakingModuleId);
-        return
-            newDepositsAllocation[stakingModuleIndex] - stakingModulesCache[stakingModuleIndex].activeValidatorsCount;
+        return Math256.min(
+            _getStakingModuleByIndex(stakingModuleIndex).maxDepositsPerBlock,
```
avsetsin, line 1053: 🤔 If we take the check for `maxDepositsPerBlock` out of the DSM we broke the semantics. The SR doesn't know anything about the limitation per block. It's a responsibility of DSM. 

In a setup where we have disconnect the DSM from the protocol, this check looks unnatural

krogla: hence we should not store maxDepositsPerBlock and minDepositBlockDistance parameters inside SR as we already have broken semantics in design.

besides, in our case the maxDepositsPerBlock parameter does not look like a dsm-related parameter, in general it looks more like a sanity check of the corresponding CL values, but in combination with the minDepositBlockDistance parameter, which is now also stored inside SR as a module parameter, it is 
more like levers to regulate the rate of deposits to a particular module.

and even when we no longer need the DSM, these parameters may well remain

avsetsin:
Decided to hold off on these optimizations for now

PR #116, opened 2024-07-02
Helper script for deploying LidoLocator on upgrade

arwer13:
- [ ] the script upgrade-locator.ts
- [ ] minimal readme for upgrade scripts
- [ ] fix of logs in scratch deploy scripts

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@ foundry/cache/
 foundry/out/
 
 .env
+accounts.json
 deployed-local.json
 
 # MacOS
--- a/docs/scratch-deploy.md
+++ b/docs/scratch-deploy.md
@@ -76,12 +76,12 @@ A brief description of what's going on under the hood in the deploy script.
 
 ## Local deployment
 
-Deploys the DAO to local (http://127.0.0.1:8545) dev node (anvil, hardhat, ganache).
+Deploys the DAO to local (http://127.0.0.1:8555) dev node (anvil, hardhat, ganache).
 The deployment is done from the default test account `0xf39Fd6e51aad88F6F4ce6aB8827279cffFb92266` derived from the default mnemonic.
 Thus the node must be configured with the default test accounts derived from the mnemonic `test test test test test test test test test test test junk`.
 
 1. Run `yarn install` (get sure repo dependencies are installed)
-2. Run the node on default port 8545 (for the commands, see subsections below)
+2. Run the node on port 8555 (for the commands, see subsections below)
 3. Run the deploy script `bash scripts/scratch/dao-local-deploy.sh` from root repo directory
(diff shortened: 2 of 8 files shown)


docs/scratch-deploy.md
```diff
@@ -76,12 +76,12 @@ A brief description of what's going on under the hood in the deploy script.
 
 ## Local deployment
 
-Deploys the DAO to local (http://127.0.0.1:8545) dev node (anvil, hardhat, ganache).
+Deploys the DAO to local (http://127.0.0.1:8555) dev node (anvil, hardhat, ganache).
```
loga4, line 79: why are we changing the default port?

scripts/upgrade/deploy-locator.ts
```diff
@@ -0,0 +1,82 @@
+import { assert } from "chai";
+import { ethers } from "hardhat";
+
+import { deployImplementation, getContractAt, LoadedContract, log, readNetworkState, Sk } from "lib";
+
+const VIEW_NAMES_AND_CTOR_ARGS = [
+  // As view names on LidoLocator
+  Sk.accountingOracle,
+  Sk.depositSecurityModule,
+  "elRewardsVault",
```
loga4, line 10: maybe use only strings or only variables here?

docs/scratch-deploy.md
arwer13: To make it consistent with the rest:
- `anvil -p 8555` at the TLDR
- `export RPC_URL=${RPC_URL:="http://127.0.0.1:8555"}` in [dao-local-deploy.sh](/scripts/scratch/dao-local-deploy.sh)

Also 8555 is used instead of 8545 to reduce probability of collision with another local node running

scripts/upgrade/deploy-locator.ts
arwer13: The ideas was to highlight difference in naming in LidoLocator in the state file.
But as the state file keys are not involved actually, I agree (pushed update).

.gitignore
```diff
@@ -18,6 +18,7 @@ foundry/cache/
 foundry/out/
 
 .env
+accounts.json
```
tamtamchik, line 21: Should we add `accounts.json.example`?

arwer13: Kind of "yes", but if we planning / hoping for encrypted pk management, I think maybe we should not  making accounts.json convenient and well established?

TheDZhon: I am for TODO and working on pk management as a separate PR

tamtamchik: Yep, I agree! 👍

arwer13: Added TODO

arwer13: @mymphe Sees to be ready to merge

mymphe (approved):

PR #136, opened 2024-07-25
Accounting Oracle integration tests

vp4242:
Accounting Oracle integration tests


test/integration/accounting.ts
```diff
@@ -0,0 +1,143 @@
...
+    snapshot = await Snapshot.take();
+  });
+
+  afterEach(async () => await Snapshot.restore(snapshot));
+
+  const getEvents = (receipt: ContractTransactionReceipt, eventName: string) => {
+    return findEventsWithInterfaces(receipt, eventName, ctx.interfaces);
+  };
+
+  // const waitToNextAvailableReportTime = async (consensusContract: IHashConsensus) => {
```
tamtamchik, line 29: This function can be used from `lib/protocol/helpers` if needed

lib/protocol/discover.ts
```diff
@@ -60,8 +60,15 @@ const loadContract = async <Name extends ContractName>(name: Name, address: stri
...
  */
-const getFoundationContracts = async (locator: LoadedContract<LidoLocator>, config: ProtocolNetworkConfig) =>
-  (await batch({
+const getFoundationContracts = async (locator: LoadedContract<LidoLocator>, config: ProtocolNetworkConfig) => {
+  const loadedWithdrawalQueue = loadContract(
+    "WithdrawalQueueERC721",
+    config.get("withdrawalQueue") || (await locator.withdrawalQueue()),
+  );
+
+  const loadedWstETH = loadContract("WstETH", config.get("wstETH") || (await (await loadedWithdrawalQueue).WSTETH()));
```
tamtamchik, line 69: This is a nice catch 👍 
I do believe that wstETH contact should belong to CoreContracts, in other words, be resolvable via the LidoLocator.

lib/protocol/helpers/accounting.ts
```diff
@@ -559,12 +559,13 @@ export const submitReport = async (
     consensusVersion,
   });
 
-  log.debug("Pushing oracle report", data);
+  log.debug("Pushing oracle report 2", data);
```
tamtamchik, line 562: Don't like 1, 2, 3... Maybe update naming to something more meaningful. But it's definitely an improvement over the initial version! 👍

test/integration/accounting.ts
```diff
@@ -0,0 +1,812 @@
...
+
+    [ethHolder] = await ethers.getSigners();
+    await setBalance(ethHolder.address, ether("1000000"));
+
+    snapshot = await Snapshot.take();
+  });
+
+  afterEach(async () => await Snapshot.restore(snapshot));
+
+  const getEvents = (receipt: ContractTransactionReceipt, eventName: string) => {
```
tamtamchik, line 36: It may be replaced with `ctx.getEvents`. This is, in fact, the same function.

test/integration/accounting.ts
```diff
@@ -0,0 +1,812 @@
...
+describe("Accounting integration", () => {
+  let ctx: ProtocolContext;
+  let snapshot: string;
+  let ethHolder: HardhatEthersSigner;
+
+  beforeEach(async () => {
+    ctx = await getProtocolContext();
+
+    [ethHolder] = await ethers.getSigners();
+    await setBalance(ethHolder.address, ether("1000000"));
```
tamtamchik, line 29: This change will persist on the fork and might spoil some other tests. So it is better to move it either inside the test or to create a snapshot in the `before` block and put this balance update into the `beforeEach` block.

test/integration/accounting.ts
```diff
@@ -0,0 +1,812 @@
...
+  });
+
+  it("Should account correctly with positive CL rebase close to the limits", async () => {
+    const { lido, accountingOracle, oracleReportSanityChecker, stakingRouter } = ctx.contracts;
+
+    const annualIncreaseLimit = (await oracleReportSanityChecker.getOracleReportLimits())[2];
+    const preCLBalance = (await lido.getBeaconStat()).slice(-1)[0];
+
+    let rebaseAmount = (preCLBalance * annualIncreaseLimit * (ONE_DAY + 1n)) / (365n * ONE_DAY) / MAX_BASIS_POINTS;
+    rebaseAmount = roundToGwei(rebaseAmount);
```
tamtamchik, line 187: I believe it's important to include a detailed description of our rebase amount calculation. Future colleagues (or yourself) will appreciate the clarity.

For example.

```suggestion
    // To calculate the rebase amount close to the annual increase limit
    // we use (ONE_DAY + 1n) to slightly underperform for the daily limit
    // This ensures we're testing a scenario very close to, but not exceeding, the annual limit
    let rebaseAmount = (preCLBalance * annualIncreaseLimit * (ONE_DAY + 1n)) / (365n * ONE_DAY) / MAX_BASIS_POINTS;
    rebaseAmount = roundToGwei(rebaseAmount);

    // At this point, rebaseAmount represents a positive CL rebase that is
    // just slightly below the maximum allowed daily increase, testing the system's
    // behavior near its operational limits
```

test/integration/accounting.ts
```diff
@@ -0,0 +1,812 @@
...
+    const totalELRewardsCollectedAfter = await lido.getTotalELRewardsCollected();
+    expect(totalELRewardsCollectedBefore).to.equal(totalELRewardsCollectedAfter);
+
+    const totalPooledEtherAfter = await lido.getTotalPooledEther();
+    expect(totalPooledEtherBefore + rebaseAmount).to.equal(totalPooledEtherAfter + amountOfETHLocked);
+
+    const sharesAsFeesList = getEvents(reportTxReceipt, "TransferShares").map((e) => e.args.sharesValue);
+    let mintedSharesSum = 0n;
+
+    if (amountOfETHLocked == 0) {
```
tamtamchik, line 212: The same goes for conditions here; why should it be 0, and in what cases? I noticed that there are comments in the code tests further. Shouldn't we add comments here as well?

tamtamchik (changes requested):
I attempted to run tests on the locally deployed version of the protocol, but they failed. Can you please check? It's possible that they are missing some provisioning.

lib/protocol/discover.ts
tamtamchik: I have created and issue #163 and PR for #164, so it will be traced separately.

tamtamchik (approved):
LGTM

PR #143, opened 2024-08-07
fix: integration tests hardhat node version

tamtamchik:
Fix version for CI Hardhat node runner, `latest` version is broken.

### DIFF
--- a/.github/workflows/tests-integration.yml
+++ b/.github/workflows/tests-integration.yml
@@ -10,7 +10,7 @@ jobs:
 
     services:
       hardhat-node:
-        image: feofanov/hardhat-node
+        image: feofanov/hardhat-node:2.22.4
         ports:
           - 8545:8545
         env:


.github/workflows/tests-integration.yml
```diff
@@ -10,7 +10,7 @@ jobs:
 
     services:
       hardhat-node:
-        image: feofanov/hardhat-node
+        image: feofanov/hardhat-node:2.22.4
```
TheDZhon, line 13: but we use 2.22.6 atm :thinking:

TheDZhon (approved):
:+1:

.github/workflows/tests-integration.yml
tamtamchik: It was easier just to revert to the latest stable version for now. Will update and sync in future when move to `lidofinance` dockerhub.

TheDZhon: :heart:

PR #148, opened 2024-08-12
Add Issue forms 

mymphe:
This PR add issue forms to the repo


TheDZhon (approved):
:heart:

.github/ISSUE_TEMPLATE/CONTRACT-BUG-REPORT.yml
```diff
@@ -0,0 +1,53 @@
+name: Contract Bug Report
+description: File a bug report in a contract.
```
TheDZhon, line 2: Should refer to the SECURITY page first :warning: 

Reason: to file a bug by 3rd-party via Immunefi

TheDZhon (changes requested):

.github/ISSUE_TEMPLATE/CONTRACT-BUG-REPORT.yml
tamtamchik: It looks like it is impossible to add a link to the GitHub issue reporting interface. Documentation refers

Screenshot 2024-08-26 at 17 11 31

tamtamchik:
Using just a single text box is way easier than filling out an entire form. It lets users think less, saves time, and they can just write what they need. I really think simplicity is the way to go for reporting. But throwing in some markdown guidance could be helpful to keep the report on point and not just a random ramble.

tamtamchik (approved):

TheDZhon (approved):
Let the sunshine in

PR #149, opened 2024-08-12
Negative rebase deploy script

vp4242:
Negative rebase deploy script

### DIFF
--- a/scripts/scratch/neg-rebase-deploy.sh
+++ b/scripts/scratch/neg-rebase-deploy.sh
@@ -48,6 +48,8 @@ yarn compile
 rm -f ${NETWORK_STATE_FILE}
 cp ${NETWORK_STATE_DEFAULTS_FILE} ${NETWORK_STATE_FILE}
 
+
 yarn compile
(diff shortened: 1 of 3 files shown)


scripts/scratch/neg-rebase-deploy.sh
```diff
@@ -0,0 +1,55 @@
...
+rm -f ${NETWORK_STATE_FILE}
+cp ${NETWORK_STATE_DEFAULTS_FILE} ${NETWORK_STATE_FILE}
+
+yarn compile
+
+rm -f ${NETWORK_STATE_FILE}
+cp ${NETWORK_STATE_DEFAULTS_FILE} ${NETWORK_STATE_FILE}
+
+
+yarn compile
```
tamtamchik, line 52: Why is that twice?

vp4242: That's probably a mistake. I would close this PR most probably as it's not needed anymore. Do you think it make sense to create a scratch deploy, specific for Sepolia? In that case this PR could be modified

tamtamchik: No, I just mentioned it in case this is still useful. I'll close the PR.

tamtamchik:
Closed as it's not needed anymore.

PR #166, opened 2024-08-28
feat: allow scratch deploy usage in local integration tests 

tamtamchik:
- Refactor scratch deploy steps runner (allow to import steps for integration provisioning)
- Update scratch deploy output style a bit
- Replaced local scratch deploy acceptance test with integration tests suite
- Some new commands for running integration tests with a scratch-deploy "flavor" and the ability to trace deployments and integration tests were added.
- Updated documentation.

Image to attract attention...
Screenshot 2024-08-29 at 14 56 48


lib/scratch.ts
```diff
@@ -0,0 +1,63 @@
...
+import { ethers } from "hardhat";
+
+import { log } from "./log";
+import { resetStateFile } from "./state-file";
+
+export async function deployScratchProtocol(networkName: string): Promise<void> {
+  await resetStateFile(networkName);
+
+  const steps = [
+    "scratch/steps/00-populate-deploy-artifact-from-env",
```
arwer13, line 12: Is it possible to deduplicate the steps here and in deploy-dao.sh?

lib/contract.ts
```diff
@@ -34,17 +34,15 @@ export async function addContractHelperFields(contract: BaseContract, name: stri
...
 ) {
   if (!signer) {
     signer = await ethers.provider.getSigner();
   }
-  const result = factory.connect(address, signer as ContractRunner);
-  const factoryName = factory.name;
-  const contractName = factoryName.slice(0, factoryName.indexOf("__"));
-  return (await addContractHelperFields(result, contractName)) as unknown as LoadedContract<ContractType>;
+  const result = await ethers.getContractAt(name, address, signer);
+  return (await addContractHelperFields(result, name)) as unknown as LoadedContract<ContractType>;
```
arwer13, line 45: Just thoughts. Force-casting to the type seems even dirtier for me now. I personally see getting rid of typechain at all a nice solution, but it would require more refactoring. Sticking to the typed contract instances properly I guess would require even more refactoring. So, no call to action anyway :)

arwer13 (approved):
Love to see scratch deployment being improved and integrated, thank you! 🤩

lib/scratch.ts
tamtamchik: Removed duplication by introducing a steps JSON file that defines the steps and their order. Not sure if we need to get rid of it, because not all the commands might be required for different scratch deploys.

PR #168, opened 2024-08-30
Feat/sr 1.5 ackee audit

mkurayan:
Fixes for ackee audit. Each commit contains the fix for a single finding.

- L1: Overflow on type casting
- L2: Potential revert on underflow.
- L3: The clearNodeOperatorPenalty returns always false
- I2: Missing event on clearNodeOperatorPenalty
- I3: Typos

Note. The commit for L2 (Potential revert on underflow) also contains contract size optimization (necessary to fit contract size in the 24 KiB limit).

### DIFF
--- a/contracts/0.8.9/StakingRouter.sol
+++ b/contracts/0.8.9/StakingRouter.sol
@@ -63,6 +63,10 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
         uint256 currentNodeOpExitedValidatorsCount,
         uint256 currentNodeOpStuckValidatorsCount
     );
+    error UnexpectedFinalExitedValidatorsCount (
+        uint256 newModuleTotalExitedValidatorsCount,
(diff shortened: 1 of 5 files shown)


contracts/0.8.9/StakingRouter.sol
```diff
@@ -1479,4 +1504,19 @@ contract StakingRouter is AccessControlEnumerable, BeaconChainDepositor, Version
...
+    function _getStakingModuleSummary(IStakingModule stakingModule) internal view returns (uint256, uint256, uint256) {
+        (
+            uint256 totalExitedValidators,
+            uint256 totalDepositedValidators,
+            uint256 depositableValidatorsCount
+        ) = stakingModule.getStakingModuleSummary();
+
+        return (totalExitedValidators, totalDepositedValidators, depositableValidatorsCount);
+    }
 }
```
krogla, line 1522: could be simplified to `return stakingModule.getStakingModuleSummary();`

krogla (changes requested):
lgtm in rest

Issue #192, opened 2024-09-04
[Feature Request]: Restore storage-layour-update CI workflow


tamtamchik:
### Summary

Need to restore the https://github.com/lidofinance/lido-dao/blob/master/.github/workflows/storage-layout-update.yml pipeline.

### Problem

Updates storage reference

### Guidelines

- [X] I agree to follow [Code of Conduct](https://github.com/lidofinance/core/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read [Lido Contribution Guide](https://github.com/lidofinance/core/blob/master/CONTRIBUTING.md).

tamtamchik:
Removed from the Repovation scope.

Issue #191, opened 2024-09-04
[Feature Request]: Restore storage-layout-check CI workflow


tamtamchik:
### Summary

Need to restore the https://github.com/lidofinance/lido-dao/blob/master/.github/workflows/storage-layout-check.yml pipeline.

### Problem

`storage-layout-check`  checks the storage layout updates for the contracts against the reference one, works only for linear storage (sic!), unstructured isn't supported

### Implementation

we can remove it if there is no easy way to add unstructured storage checks

### Guidelines

- [X] I agree to follow [Code of Conduct](https://github.com/lidofinance/core/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read [Lido Contribution Guide](https://github.com/lidofinance/core/blob/master/CONTRIBUTING.md).

tamtamchik:
Removed from the Repovation scope.

PR #195, opened 2024-09-04
chore: restore deployed files

tamtamchik:
Restore `deployed-*` artifacts from the https://github.com/lidofinance/lido-dao repo

## Context

Those files might be required for new deploys and verifications lidofinance/lido-dao#848, lidofinance/lido-dao#847, lidofinance/core#191, lidofinance/core#192 and others


TheDZhon:
wdyt, @arwer13?

tamtamchik:
Should be moved to https://github.com/lidofinance/scripts repo?

PR #209, opened 2024-09-11
feat: add target limit mode check

mkurayan:
Check the target limit mode value in the nor contract


contracts/0.4.24/nos/NodeOperatorsRegistry.sol
```diff
@@ -109,6 +109,8 @@ contract NodeOperatorsRegistry is AragonApp, Versioned {
     uint8 internal constant TARGET_VALIDATORS_COUNT_OFFSET = 1;
     /// @dev actual operators's number of keys which could be deposited
     uint8 internal constant MAX_VALIDATORS_COUNT_OFFSET = 2;
+    /// @dev Forced target limit mode
+    uint8 internal constant FORCED_TARGET_LIMIT_MODE = 2;
```
avsetsin, line 113: In this implementation, we will not be able to add a new mode in VEBO without contract changes. We discussed that we could use an immutable variable for the constraint.

PR #217, opened 2024-09-24
ci: use lidofinance/hardhat-node as a base node for CI actions

tamtamchik:
Use ghcr.io/lidofinance/hardhat-node as node runner for CI


rnmsslido (approved):
LGTM

PR #220, opened 2024-09-24
feat: add hardhat task to verify deployed contracts

tamtamchik:
Screenshot 2024-09-24 at 23 13 10


tamtamchik:
@TheDZhon @mymphe, any objections against coupling this logic (abi extraction, contracts verification) with HardHat tasks are welcome; I'm not rock solid about this solution. It really might be that old plain bash files work better here.

tamtamchik:
Proof of work: https://sepolia.etherscan.io/address/0x34104f0b5bf1ad71ec465240e3e8fe1547dac0e6

tasks/verify-contracts.ts
```diff
@@ -0,0 +1,181 @@
...
+function getDeployedContracts(networkState: NetworkState): DeployedContract[] {
+  return Object.values(networkState)
+    .filter((contract): contract is Contract => typeof contract === "object")
+    .flatMap(getDeployedContract);
+}
+
+async function verifyContractList(contracts: DeployedContract[], hre: HardhatRuntimeEnvironment) {
+  for (const contract of contracts) {
+    await verifyContract(contract, hre);
+  }
```
mymphe, line 84: We could parallelize this using Promise.all. Could speed up verification process if there is a lot of contracts to verify.

mymphe (approved):
Suggested minor improvements

globals.d.ts
```diff
@@ -66,5 +66,8 @@ declare namespace NodeJS {
     MAINNET_VALIDATORS_EXIT_BUS_ORACLE_ADDRESS?: string;
     MAINNET_WITHDRAWAL_QUEUE_ADDRESS?: string;
     MAINNET_WITHDRAWAL_VAULT_ADDRESS?: string;
+
+    /* for contract verification */
```
TheDZhon, line 70:
```suggestion
    /* for contract sourcecode verification with `hardhat-verify` */
```

tasks/extract-abis.ts
```diff
@@ -0,0 +1,41 @@
...
+  "@aragon/apps-vault/contracts/Vault.sol:Vault",
+  "@aragon/apps-lido/apps/voting/contracts/Voting.sol:Voting",
+  "@aragon/apps-lido/apps/token-manager/contracts/TokenManager.sol:TokenManager",
+];
+const SKIP_NAMES_REGEX = /(Mock|Harness|test_helpers|Imports|deposit_contract|Pausable|.dbg.json|build-info)/;
+
+task("abis:extract", "Extract ABIs from artifacts").setAction(async (_: unknown, hre: HardhatRuntimeEnvironment) => {
+  await hre.run("compile");
+
+  const artifactNames = await hre.artifacts.getAllFullyQualifiedNames();
```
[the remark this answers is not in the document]
TheDZhon: yeah, good catch

tasks/extract-abis.ts
```diff
@@ -0,0 +1,41 @@
...
+
+import { log, yl } from "lib/log";
+
+const ABI_OUTPUT_PATH = path.resolve(process.cwd(), "lib", "abi");
+const LIDO_ARTIFACT_PREFIX = "contracts/";
+const ARAGON_ARTIFACT_PATHS = [
+  "@aragon/apps-finance/contracts/Finance.sol:Finance",
+  "@aragon/apps-vault/contracts/Vault.sol:Vault",
+  "@aragon/apps-lido/apps/voting/contracts/Voting.sol:Voting",
+  "@aragon/apps-lido/apps/token-manager/contracts/TokenManager.sol:TokenManager",
```
TheDZhon, line 15: later on we might vendor these deps tbh, too obsolete and orphaned as of now

tasks/verify-contracts.ts
```diff
@@ -0,0 +1,181 @@
...
+
+  const networkState = await getNetworkState(network, deployerAddress);
+  const deployedContracts = getDeployedContracts(networkState);
+  await verifyContractList(deployedContracts, hre);
+}
+
+async function getDeployerAddress(hre: HardhatRuntimeEnvironment): Promise<string> {
+  const deployer = await hre.ethers.provider.getSigner();
+  if (!deployer) {
+    throw new Error("Deployer not found!");
```
TheDZhon, line 57: why do we need finding a deployer at all though?

tasks/verify-contracts.ts
```diff
@@ -0,0 +1,181 @@
...
+}
+
+async function getNetworkState(network: string, deployerAddress: string): Promise<NetworkState> {
+  const networkStateFile = `deployed-${network}.json`;
+  const networkStateFilePath = path.resolve("./", networkStateFile);
+  const data = await fs.readFile(networkStateFilePath, "utf8");
+  const networkState = JSON.parse(data) as NetworkState;
+
+  if (networkState.deployer !== deployerAddress) {
+    throw new Error(`Deployer address mismatch: ${networkState.deployer} != ${deployerAddress}`);
```
TheDZhon, line 69: not sure if it's needed here

tasks/verify-contracts.ts
```diff
@@ -0,0 +1,181 @@
...
+function getDeployedContract(contract: Contract): DeployedContract[] {
+  if ("proxy" in contract && "implementation" in contract) {
+    return [contract.proxy, contract.implementation];
+  } else if ("contract" in contract && "address" in contract && "constructorArgs" in contract) {
+    return [contract];
+  }
+  return [];
+}
+
+async function isContractVerified(address: string, hre: HardhatRuntimeEnvironment): Promise<boolean> {
```
TheDZhon, line 126: Can't we re-use
https://github.com/NomicFoundation/hardhat/blob/main/packages/hardhat-verify/src/internal/tasks/etherscan.ts#L95-L108 ?

tasks/verify-contracts.ts
```diff
@@ -0,0 +1,181 @@
...
+  if (typeof apiKey === "string") {
+    return apiKey;
+  } else if (typeof apiKey === "object" && apiKey !== null) {
+    return apiKey[hre.network.name] || "";
+  }
+  return "";
+}
+
+async function fetchContractSourceCode(apiURL: string, params: URLSearchParams) {
+  return await retryFetch(`${apiURL}?${params}`, MAX_RETRY_ATTEMPTS, RETRY_DELAY_MS);
```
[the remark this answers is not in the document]
TheDZhon: should have been caught by the linter, hmm

TheDZhon (approved):
I am totally ok with the hardhat verification approach given that it's totally automated and strongly typed against the previous bash solution :muscle: 

Left some questions here and there :pray:

tasks/extract-abis.ts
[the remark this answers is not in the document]
tamtamchik: I saw that `compile` actually includes a subtask called `compile:remove-obsolete-artifacts` which should handle the cleanup. But I'm not against the separate `cleanup` step.

tasks/verify-contracts.ts
tamtamchik: It'll lead to bumping into rate limits and logs mess, and considering the number of contracts and rare occasions for this action, I decided to sacrifice speed here :)

tasks/verify-contracts.ts
tamtamchik: Removed. Thanks!

tasks/verify-contracts.ts
tamtamchik: Yep, missed that, thank you!

PR #225, opened 2024-10-02
chore: repovation update

tamtamchik:
- #220 
- #224


tamtamchik:
@rnmsslido, we have a failing workflow here; I'm surprised that you didn't question this. 
The fix is in separate PR https://github.com/lidofinance/core/pull/224 (also waiting for review)

rnmsslido:
> @rnmsslido, we have a failing workflow here; I'm surprised that you didn't question this. The fix is in separate PR #224 (also waiting for review)

Only change I could review in this PR was on `.github/workflows/release-abis.yml`, which seemed like a simple change.

As for the failed workflow, it was related to code coverage and we don't maintain that code (if you need help from us debugging the error, it's another story: you can create a task for us on linear and contact your devops liaison).

If the PR isn't ready for review for whatever reason, please mark it as a draft. I'll check #224 shortly.

tamtamchik:
@rnmsslido roger, thank you for the clarification! 🫡

PR #229, opened 2024-10-09
backport: changes from master to develop

tamtamchik:
Want to prepare the develop branch to be ported to lido-dao.


.github/workflows/coverage.yml
```diff
@@ -7,6 +7,11 @@ jobs:
     name: Hardhat
     runs-on: ubuntu-latest
 
+    permissions:
```
rnmsslido, line 10: Now I see these permissions are maybe too open; have you tested it with only `contents: read`?
If no actions are using PRs and issues, it should be enough.

tamtamchik: Changed in 17c5ca7390d3038287bb3d9d7424aadac0922e3f

TheDZhon (approved):
image

.github/workflows/coverage.yml
tamtamchik: Well, it pushes to the coverage reports branch (to be able to make diffs), so it needs the write permissions.

TheDZhon:
image

.github/workflows/coverage.yml
rnmsslido: I see, in this case `contents: write` is needed indeed. These ones could be removed, then?
```yaml
      issues: write
      pull-requests: write
 ```
Should be a good test to do on `develop`.

